PR #36, opened 2020-12-15
Default save path pref to safe directory

stefcameron:
Relies on https://github.com/lensapp/lens/pull/1785 to work, which is in Lens 4.0.4.

Lens now generates data directories for extensions that are safe,
even if the extension gets uninstalled. This commit uses that
generated directory as the new default save path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # ChangeLog
 
+## UNRELEASED
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


stefcameron:
Waiting to hear on https://github.com/lensapp/lens/pull/1785 whether the directory is cleaned-up on extension removal or not. If it is, then we'll have to default to the user's home directory instead.

PR #205, opened 2021-03-22
Wrapping emotion instance into CacheProvider

aleksfront:
Lens is using architecture when multiple emotion instances can be activated. Currently, all of them shares same `<style>` tag causing dropping all of the styles inside of it. Original bug: https://github.com/lensapp/lens/issues/2219.

This PR wraps all `emotion` contents into `<CacheProvider/>` which helps to isolate MCC Extension styles and prevents other emotion styles to be removed.

Part of https://github.com/lensapp/lens/pull/2365

Signed-off-by: Alex Andreev <alex.andreev.email@gmail.com>


stefcameron (approved):
This is a great find, thank you!

PR #622, opened 2021-12-01
[PRODX-19642] AuthAccess to Cloud

AntonNos:
Renamed AuthAccess to Cloud, added new property cloudUrl and dropped usesSso property


stefcameron:
👀

src/renderer/auth/Cloud.js
```diff
@@ -54,9 +54,8 @@ export class AuthAccess {
     // IDP client associated with current tokens; undefined if unknown; null if not specified
     idpClientId: [rtv.OPTIONAL, rtv.STRING],
 
-    // true if access is via SSO; false if it's via basic auth (which would require a
-    //  password, and which we no longer support); null if it's undetermined
-    usesSso: [rtv.OPTIONAL, rtv.BOOLEAN],
+    //
```
stefcameron, line 57: Missing a comment here, like "URL to the MCC instance"

src/renderer/auth/Cloud.js
```diff
@@ -11,9 +11,9 @@ const hasPassed = function (date) {
 
 /**
  * Captures API authorization tokens and expiries.
```
stefcameron, line 13: I think we should update this JSDoc comment to more clearly identify what this class now represents, something like,

```
Represents a connection to a single MCC instance (i.e. Management Cluster),
including access tokens and expiry times.
```

lens-extension-cc
```diff
@@ -0,0 +1 @@
+/Users/anos/.k8slens/extensions/lens-extension-cc
```
stefcameron, line 1: This shouldn't be included in the PR. Looks like a file to be deleted.

src/renderer/components/GlobalPage/Login.js
```diff
@@ -189,15 +185,17 @@ export const Login = function () {
...
-            `${strings.login.error.basicAuth()} ${strings.noteOwner}`
-          );
-        }
+        // if (usesSso) {
+        //   startLogin();
+        // } else {
+        //   Notifications.error(
+        //     `${strings.login.error.basicAuth()} ${strings.noteOwner}`
+        //   );
+        // }
```
stefcameron, line 198: Let's delete this old commented code.

stefcameron (changes requested):
The auto-login when the extension is loaded is actually happening because of code in useClusterLoader.js, not the effect in Login.js. If you've already connected once before, then we have the URL, and on load, we attempt to get the config, which succeeds, and then we see that we have an invalid Cloud object, so we assume we should connect to get tokens: https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer/hooks/useClusterLoader.js#L127

But I think that's OK. We're going to make many other changes to the whole connection process that will quickly make this problem irrelevant anyway. So it's fine for now.

src/renderer/components/GlobalPage/Login.js
AntonNos: I commented it because I thought that we still need Notifications.error, but not sure where should I place it.

stefcameron: It'll change completely, so we can drop it for now. I'll ask Michal to add this error case to the mocks.

src/renderer/auth/Cloud.js
AntonNos: Comment added

src/renderer/auth/Cloud.js
AntonNos: Updated

lens-extension-cc
AntonNos: File was deleted

src/renderer/components/GlobalPage/Login.js
AntonNos: Deleted

src/renderer/auth/Cloud.js
```diff
@@ -54,9 +55,8 @@ export class AuthAccess {
     // IDP client associated with current tokens; undefined if unknown; null if not specified
     idpClientId: [rtv.OPTIONAL, rtv.STRING],
 
-    // true if access is via SSO; false if it's via basic auth (which would require a
-    //  password, and which we no longer support); null if it's undetermined
-    usesSso: [rtv.OPTIONAL, rtv.BOOLEAN],
+    // URL to the MCC instance
+    cloudUrl: [rtv.OPTIONAL, rtv.STRING],
```
stefcameron, line 59: One last thing I just noticed: This should be
```suggestion
    cloudUrl: [rtv.EXPECTED, rtv.STRING],
```
because the property is no longer optional. We want all Cloud instances to have a `cloudUrl`. But then you need to be able to `new Cloud()` to create one where `cloudUrl` will be null, initially, so in terms of [RTV.js](https://rtvjs.stefcameron.com/) validations, that means the value is `EXPECTED`. (RTV.js is a runtime validation library I wrote.)

src/renderer/auth/Cloud.js
```diff
@@ -226,31 +222,30 @@ export class AuthAccess {
...
      */
-    Object.defineProperty(this, 'usesSso', {
+    Object.defineProperty(this, 'cloudUrl', {
       enumerable: true,
       get() {
-        return _usesSso;
+        return _cloudUrl;
       },
       set(newValue) {
         const normalizedNewValue = newValue ?? null; // undefined -> null
```
stefcameron, line 241: And then this should change to
```suggestion
        const normalizedNewValue = newValue || null; // '' or undefined -> null
```
so that we also normalize empty strings to null since an empty string means nothing here. It's either null or a non-empty string. Before, it was `usesSso` and it was a boolean, which had different normalization requirements because a boolean is either true or false (and if not set, we'd normalize to null).

stefcameron (commented):
Thanks for fixes. Just 2 more small things.

src/renderer/auth/Cloud.js
AntonNos: fixed

src/renderer/auth/Cloud.js
AntonNos: fixed

stefcameron (approved):
Thanks for the fixes, good to merge!

Note that when you merge this, you'll need to rebase your local "PR2" branch onto master and then force-push the result to your "PR2" branch in order to get it updated to just the "PR2" changes. Otherwise, GitHub will continue to show "PR2" as being all the changes from "PR1" and "PR2". LMK if you have any questions about that.

PR #628, opened 2021-12-08
Adding SshKey, Credential, Proxy Catalog entities

stefcameron:
Adds custom Catalog categories and entities for SSH Keys, Credentials, and Proxies.

Fake items are generated in the Catalog in Dev builds only, just to show how it all works
and provide a template for future work when we'll need these entities (and possibly
others).

Each entity has a custom catalog menu, context menu, and details panel.

Also the native cluster category gets a custom catalog menu item for
creating new clusters.

None the actions do anything other than log to the console for now. We'll
fill this in once we get to implementing those features.

### Previews

Custom categories, entities, and catalog menus:

image

Custom context menus:

image

Customized Details view ("Context menu > View details"):

image

Extended the Catalog menu for clusters:

image


src/catalog/SshKeyCatalogEntity.ts
```diff
@@ -0,0 +1,137 @@
...
+        name: consts.catalog.entities.sshKey.versions.v1alpha1,
+        entityClass: SshKeyEntity,
+      },
+    ],
+    names: {
+      kind: SshKeyEntity.kind,
+    },
+  };
+
+  // DEBUG TODO: what is this? it's being called below; I don't recognize this syntax,
```
Nokel81, line 113: This is a optional static field of type function that takes no args and produces `void`.

This was used in for weblinks because of the decision to have the same class be used both on renderer and main but the `onAdd` code is only applicable on `renderer` so we initialize this static field to a function there.

src/catalog/SshKeyCatalogEntity.ts
```diff
@@ -0,0 +1,137 @@
...
+  //  Is it TypeScript or some new ECMAScript proposal?
+  public static onAdd?: () => void;
+
+  constructor() {
+    super();
+
+    // DEBUG NOTE: this must be the big blue "+" button
+    // DEBUG TODO: there's a type error with `this.on` which seems wrong, but why?
```
Nokel81, line 121: what is the type error?

src/catalog/SshKeyCatalogEntity.ts
```diff
@@ -0,0 +1,137 @@
...
+          SshKeyCategory.onAdd();
+        },
+      });
+    });
+  }
+}
+
+// DEBUG TODO: Can I actually add to both right here?
+Renderer.Catalog.catalogCategories.add(new SshKeyCategory());
+Main.Catalog.catalogCategories.add(new SshKeyCategory());
```
Nokel81, line 137: yes but I would check to make sure that it is defined first

src/renderer/renderer.tsx
```diff
@@ -88,6 +90,25 @@ export default class ExtensionRenderer extends LensExtension {
     },
   ];
 
+  catalogEntityDetailItems = [
+    {
+      kind: SshKeyEntity.kind,
+      apiVersions: [SshKeyEntity.apiVersion],
+      // DEBUG TODO: what is priority? what does it do? do I need it?
+      // priority: 10,
```
Nokel81, line 98: It is used for sorting the detail items.

src/catalog/SshKeyCatalogEntity.ts
```diff
@@ -0,0 +1,137 @@
+import {
+  Renderer,
+  Main,
+  // DEBUG TODO: the TSC complains that none of the types below are known, yet they are
```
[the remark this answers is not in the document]
stefcameron: But isn't this the "top level"? How do I import this stuff? Or am I supposed to
```
import { Renderer } from '@k8slens/extensions';

...

export interface SshKeyStatus extends Renderer.CatalogEntityStatus {
  phase: SshKeyStatusPhase;
}
```

Or am I just not supposed to use these types? I don't like types anyway. 😉

src/catalog/SshKeyCatalogEntity.ts
```diff
@@ -0,0 +1,137 @@
...
+  CatalogCategory,
+  CatalogEntity,
+  CatalogEntityAddMenuContext,
+  CatalogEntityContextMenuContext,
+  CatalogEntityMetadata,
+  CatalogEntityStatus,
+} from '@k8slens/extensions';
+import * as consts from '../constants';
+import { logger as loggerUtil } from '../util/logger';
+// DEBUG TODO: import { WeblinkStore } from '../weblink-store';
```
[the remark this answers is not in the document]
stefcameron: My bad; shouldn't have been a TODO.

src/catalog/SshKeyCatalogEntity.ts
```diff
@@ -0,0 +1,137 @@
...
+  SshKeySpec
+> {
+  public static readonly apiVersion = `${consts.catalog.entities.sshKey.group}/${consts.catalog.entities.sshKey.versions.v1alpha1}`;
+  public static readonly kind = consts.catalog.entities.sshKey.kind;
+
+  public readonly apiVersion = SshKeyEntity.apiVersion;
+  public readonly kind = SshKeyEntity.kind;
+
+  // "runs" the entity; called when the user just clicks on the item in the Catalog
+  async onRun() {
```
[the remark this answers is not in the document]
stefcameron: Thanks, I see that in node_modules/@k8slens/extensions/dist/src/common/catalog/catalog-entity.d.ts now.

src/catalog/SshKeyCatalogEntity.ts
```diff
@@ -0,0 +1,137 @@
...
+    // DEBUG window.open(this.spec.url, '_blank');
+    // DEBUG TODO: somehow trigger showing details panel if possible
```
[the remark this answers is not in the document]
stefcameron: Right, but how would you "run" an SSH Key object? The logical thing to do seems to be to trigger the details panel to appear in this case. "Running" a web link by opening it makes sense. Or I guess we can ignore onRun() and just do nothing.

src/catalog/SshKeyCatalogEntity.ts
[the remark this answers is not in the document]
Nokel81: You are supposed to do that second item. This is what we do https://github.com/lensapp/lens/blob/master/extensions/pod-menu/src/logs-menu.tsx#L25-L40 since TypeScript doesn't currently support type destructuring of namespaces.

src/catalog/SshKeyCatalogEntity.ts
stefcameron: It's that "on" is not a valid member: `Property 'on' does not exist on type 'SshKeyCategory'.`

Same error with this code here, with `emit()`:
```
    Renderer.Catalog.catalogCategories
      .getCategoryForEntity<SshKeyCategory>(this)
      ?.emit('contextMenuOpen', this, context);
```

Something to do with the `TypeEmitter` package, I think.

src/catalog/SshKeyCatalogEntity.ts
stefcameron: What do you mean by checking if it's defined? Do you mean `Main/Renderer.Catalog.catalogCategories`? Both are defined from an "Intellisense" standpoint. I guess we'll see when I finally get the code to compile.

src/catalog/SshKeyCatalogEntity.ts
Nokel81: Maybe... but `@k8slens/extensions` has that package as a dependency...

src/catalog/SshKeyCatalogEntity.ts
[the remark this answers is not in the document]
stefcameron: I see. I assume all these types could be under `Renderer.Catalog`, then, but none of them are there:
image

TCS continues to complain:
image

This is what appears to be available to me in Renderer:
image

`Catalog` seems like the obvious place... Is it possible these types aren't exported correctly?

src/catalog/SshKeyCatalogEntity.ts
stefcameron: I know, and that package is installed thanks to the dependency as well.

src/catalog/SshKeyCatalogEntity.ts
Nokel81: Yes `Renderer` and `Main` will only be defined on their respective sides.

src/catalog/SshKeyCatalogEntity.ts
Nokel81: I couldn't reproduce this...

stefcameron: Thanks for trying, I appreciate it. I'll use the escape hatch to get around it. I have no idea what's going on, and not motivated to fight with the TSC more than I really have to.

```
(this as any).on(...)

    const category: any = Renderer.Catalog.catalogCategories
      .getCategoryForEntity<SshKeyCategory>(this);
    category?.emit('contextMenuOpen', this, context);
```

src/catalog/SshKeyCatalogEntity.ts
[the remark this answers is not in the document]
Nokel81: They are under `Common.Catalog`

src/catalog/SshKeyCatalogEntity.ts
stefcameron: I discovered the root of the issue: I was importing the class as a type instead of a class object from which to extend. Once I fixed that, the type error went away (as did many others).

src/catalog/SshKeyCatalogEntity.ts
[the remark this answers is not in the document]
stefcameron: That did it, thank you!

src/catalog/CredentialEntity.ts
```diff
@@ -0,0 +1,129 @@
...
+        icon: 'launch', // NOTE: must be a string; cannot be a component that renders an icon
+        onClick: async () => {
+          logger.log(
+            'CredentialEntity/CredentialEntity.onContextMenuOpen.browserOpen',
+            'opening Credential %s in browser...',
+            this.metadata.name
+          );
+        },
+        // can optionally show a confirmation dialog
+        // confirm: {
```
AntonNos, line 67: Looks like we don't use this commented code. Maybe need to delete it?

src/catalog/ProjectEntity.ts
```diff
@@ -0,0 +1,129 @@
...
+        icon: 'launch', // NOTE: must be a string; cannot be a component that renders an icon
+        onClick: async () => {
+          logger.log(
+            'CredentialEntity/CredentialEntity.onContextMenuOpen.browserOpen',
+            'opening Credential %s in browser...',
+            this.metadata.name
+          );
+        },
+        // can optionally show a confirmation dialog
+        // confirm: {
```
AntonNos, line 67: Same here? Or this code commented for the future updates?

src/catalog/ProxyEntity.ts
```diff
@@ -0,0 +1,127 @@
...
+        icon: 'launch', // NOTE: must be a string; cannot be a component that renders an icon
+        onClick: async () => {
+          logger.log(
+            'ProxyEntity/ProxyEntity.onContextMenuOpen.browserOpen',
+            'opening Proxy %s in browser...',
+            this.metadata.name
+          );
+        },
+        // can optionally show a confirmation dialog
+        // confirm: {
```
AntonNos, line 65: Also here, commented  code should be deleted?

src/catalog/SshKeyEntity.ts
```diff
@@ -0,0 +1,125 @@
...
+        icon: 'launch', // NOTE: must be a string; cannot be a component that renders an icon
+        onClick: async () => {
+          logger.log(
+            'SshKeyEntity/SshKeyEntity.onContextMenuOpen.browserOpen',
+            'opening SSH key %s in browser...',
+            this.metadata.name
+          );
+        },
+        // can optionally show a confirmation dialog
+        // confirm: {
```
AntonNos, line 63: Same for previous comments?

AntonNos (commented):
Tried it locally, looks great. Thank you! 
I left 4 comments about commented code, but maybe them for future update?

stefcameron:
> Tried it locally, looks great. Thank you! I left 4 comments about commented code, but maybe them for future update?

Thanks for checking it, and thanks for trying it locally. I removed the commented code. I meant to show what features were available, but I think it's more confusing than beneficial because it just looks like commented code.

stefcameron:
I also removed the ProjectEntity. It's not hooked up to anything, and it's not yet clear if we'll have an entity for a project.

PR #630, opened 2021-12-08
[PRODX-19642]: added ssoUtil

AntonNos:


src/util/ssoUtil.js
```diff
@@ -0,0 +1,74 @@
...
+    // no code, something went wrong
+    error = oAuth.error || oAuth.error_description || 'unknown';
+  }
+
+  if (error) {
+    logger.error(
+      'ssoUtil.finishAuthorization()',
+      `Failed to get tokens from authorization code, error="${error}"`
+    );
+    throw new Error(strings.ssoAuthProvider.error.authCode());
```
stefcameron, line 55: Should just throw the original error here (I admit it would be less confusing to rename the local `error` variable to `errorMsg` so it's more obvious it's a string, not an Error object):
```suggestion
throw new Error(error);
```

src/renderer/store/SsoAuthProvider.js
```diff
@@ -142,8 +46,19 @@ export const useSsoAuth = function () {
...
+          pr.reset(true);
+
           // must not be loading (we can only handle one load at a time)
-          _startAuthorization(options);
+          try {
+            ssoUtil.startAuthorization(options);
+          } catch (error) {
+            pr.loading = false;
+            pr.loaded = true;
+            pr.error = error;
```
stefcameron, line 57: As I mentioned on the original PR, `error` is an object here, while `pr.error` is a string, and it's user-facing, so this should be:
```suggestion
pr.error = strings.ssoAuthProvider.error.basicOnly();
```

src/util/ssoUtil.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {Object} options.config MCC Config object.
+ */
+const startAuthorization = function ({ config }) {
+  const authClient = new AuthClient({ config });
+
+  if (config.keycloakLogin) {
+    const url = authClient.getSsoAuthUrl();
+    Util.openExternal(url); // open in default browser
+  } else {
+    throw new Error(strings.ssoAuthProvider.error.basicOnly());
```
stefcameron, line 23: Since this is a utility module, we don't need to use a user-facing string for this:
```suggestion
throw new Error('Cloud instance does not support SSO');
```

src/util/ssoUtil.js
```diff
@@ -0,0 +1,74 @@
...
+      );
+      throw new Error('Failed to get username from token JWT');
+    }
+  }
+};
+
+export const ssoUtil = {
+  startAuthorization,
+  finishAuthorization,
+};
```
stefcameron, line 74: Instead of this, let's just export each function directly. This is more like the CJS style where we'd do:
```javascript
module.exports = {
  startAuthorization,
  finishAuthorization,
}
```

src/util/ssoUtil.js
```diff
@@ -0,0 +1,74 @@
...
+ *  for access tokens.
+ *
+ * @param {Object} options
+ * @param {Object} options.oAuth OAuth response data from request for auth code.
+ *  See `extEventOauthCodeTs` typeset in `eventBus.ts` for expected shape.
+ * @param {Object} options.config MCC Config object.
+ * @param {Cloud} options.cloud Current authentication information.
+ *  This instance WILL be cleared and updated with new tokens.
+ */
+const finishAuthorization = async function ({ oAuth, config, cloud }) {
```
stefcameron, line 38:
```suggestion
export const finishAuthorization = ...
```

src/util/ssoUtil.js
```diff
@@ -0,0 +1,74 @@
...
+const { Util } = Common;
+
+/**
+ *  Start authorization with MCC to get the temp access code via the
+ *  redirect URI that will use the 'lens://' protocol to redirect the user
+ *  to Lens and ultimately call back into `finishAuthorization()`.
+ * @param {Object} options
+ * @param {Object} options.config MCC Config object.
+ */
+const startAuthorization = function ({ config }) {
```
stefcameron, line 16:
```suggestion
export const startAuthorization = ...
```

src/renderer/auth/Cloud.js
```diff
@@ -56,7 +56,7 @@ export class Cloud {
     idpClientId: [rtv.OPTIONAL, rtv.STRING],
 
     // URL to the MCC instance
-    cloudUrl: [rtv.EXPECTED, rtv.STRING],
+    cloudUrl: [rtv.OPTIONAL, rtv.STRING],
```
stefcameron, line 59: Should back this out. It should be EXPECTED...

src/renderer/auth/Cloud.js
```diff
@@ -238,7 +238,7 @@ export class Cloud {
         return _cloudUrl;
       },
       set(newValue) {
-        const normalizedNewValue = newValue || null; // '' or undefined -> null
+        const normalizedNewValue = newValue && null; // '' or undefined -> null
```
stefcameron, line 241: Should back this out. The `||` is the right way to do it because an empty string will become `null`, which is what we want. The property is either a non-empty string, or it's null. An empty string is not a possibility. Hence, we "normalize" any falsy values to null.

stefcameron (changes requested):

src/util/ssoUtil.js
AntonNos: Exported this modules without `module.exports`, because received this error:
[EXTENSIONS-LOADER]: can't load renderer for "@mirantis/lens-extension-cc": Error: ES Modules may not assign module.exports or exports.*, Use ESM export syntax, instead: ./src/util/ssoUtil.js

stefcameron: Oops, looks like I didn't finish my though on my original comment for some reason. 🤷‍♂️ I must've seen a squirrel outside or something...

What I meant to say was that what you're doing here _is like_ the CJS `module.exports = {...}` style, but with ESM, we can export things directly, so _instead_ of this, just use
```javascript
export const startAuthorization = function (...) {...};

export const finishAuthorization = function (...) {...};
```

stefcameron: That will allow importing in different ways:
```javascript
import { startAuthorization } from './ssoUtil';

startAuthorization(...);

---

import * as ssoUtil from './ssoUtil';

ssoUtil.startAuthorization(...);
```

If you do it the way you're doing it, then the second import above wouldn't work well. You'd end-up with `ssoUtil.ssoUtil.startAuthorization()`, and the first import would fail because we're only exporting one symbol, which is an object named `ssoUtil`.

stefcameron: Looks like you understood what I was saying after all. 😄

stefcameron (changes requested):
The code looks OK, but we still have the problem that after these changes, the extension no longer works -- and these changes should not cause anything to break. I don't get any clusters back anymore. This is what I see after doing the SSO sign in:

image

stefcameron:
I should be seeing the list of available clusters...

stefcameron:
This is what we should see after SSO:

image

src/renderer/store/SsoAuthProvider.js
```diff
@@ -159,7 +74,16 @@ export const useSsoAuth = function () {
       finishAuthorization(options) {
         if (pr.loading) {
           // must be loading (i.e. in the middle of the process)
-          _finishAuthorization(options);
+          try {
+            ssoUtil.finishAuthorization(options);
```
stefcameron, line 78: I think this is the problem why it's not working: We're not `await`-ing this async function...

stefcameron: Hmm, but we weren't awaiting it before... OK, maybe that's not it.

stefcameron (approved):
This looks good! Thanks for fixing that elusive bug!

PR #646, opened 2021-12-22
PRODX-20521 Update handleProtocolActivateCluster

bnalisnikovskiy:
Changed logic at `renderer.tsx` related to handleProtocolActivateCluster
Instead of redirect to GlobalPage (SyncView) we navigate to the cluster

### DIFF
--- a/src/renderer/components/GlobalPage/SyncView.js
+++ b/src/renderer/components/GlobalPage/SyncView.js
@@ -33,6 +33,8 @@ import {
   extEventAddClustersTs,
   extEventKubeconfigTs,
   extEventOauthCodeTs,
+  EXT_EXTERNAL_ERROR_SHOW,
+  extExternalErrorShowTs,
   addExtEventHandler,
   removeExtEventHandler,
 } from '../../eventBus';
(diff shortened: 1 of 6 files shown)


stefcameron:
👀

src/renderer/renderer.tsx
```diff
@@ -157,17 +161,28 @@ export default class ExtensionRenderer extends LensExtension {
...
+      // During new design implementation we find a proper place fot this
+      this.navigate(ROUTE_GLOBAL_PAGE);
+      dispatchExtEvent({
+        type: EXT_EXTERNAL_ERROR_SHOW,
+        data: {
+          error: strings.renderer.clusterActions.error.clusterNotFound(
+            `${namespace}/${clusterName}`
+          ),
+        },
+      });
```
stefcameron, line 184: Should just be `Renderer.Components.Notifications.error(...)`. That will show a notification "toast" message in the upper-right corner of the Lens UI.

src/renderer/components/GlobalPage/SyncView.js
```diff
@@ -250,45 +248,22 @@ export const SyncView = function () {
...
+  // so probably we remove this part and show only logger
+  // at renderer.tsx. When refactoring will be finished
+  // we return this(or similar) notification logic.
+  const handleExternalErrorEvent = useCallback(function (event) {
+    const results = rtv.check({ event }, { event: extExternalErrorShowTs });
+    if (results.valid && event?.data?.error) {
+      setActiveEventType(EXT_EXTERNAL_ERROR_SHOW);
+      setExtEventError(event.data.error);
+    }
+  }, []);
```
stefcameron, line 266: This shouldn't be necessary as we'll just show the error notification from `renderer.tsx` (see my other comment there).

stefcameron (changes requested):
Nice clean-up work! I don't think we need the `EXT_EXTERNAL_ERROR_SHOW` event, though...

src/renderer/renderer.tsx
bnalisnikovskiy: Thanks. I was looking something like this. Much better now!

src/renderer/components/GlobalPage/SyncView.js
bnalisnikovskiy: Fixed

stefcameron (approved):
Looks good, thanks! I tested it locally, works in success and error cases. 👌

PR #662, opened 2022-01-05
PRODX-19969 Remove redundant Add Cluster/config logic

bnalisnikovskiy:


stefcameron (changes requested):
There's something about deleting code that can be therapeutic, right? 😌

The README also mentions [add one cluster](https://github.com/Mirantis/lens-extension-cc/blob/master/README.md#protocol---add-one-cluster) and [add multiple clusters](https://github.com/Mirantis/lens-extension-cc/blob/master/README.md#protocol---add-multiple-clusters).

Let's clean that up too as part of this since we won't support those `lens:` protocol requests anymore.

bnalisnikovskiy:
Yeah, the better type of tickets :)
Fixed, thanks

stefcameron (approved):

PR #663, opened 2022-01-05
[PRODX-20592]: Added new Welcome page (UI Part)

AntonNos:
Screenshot 2022-01-05 at 17 18 41
Logic will be added in separate PRs


src/renderer/components/GlobalPage/WelcomeView.js
```diff
@@ -0,0 +1,145 @@
...
+    fontSize: 36,
+    lineHeight: '40px',
+    textAlign: 'center',
+    marginBottom: 24,
+  };
+});
+
+const WelcomeDescription = styled.p(function () {
+  return {
+    marginBottom: 36,
```
bnalisnikovskiy, line 72: I think better use `layout.gap, .pad, .grid` as a base for margin and paddings.

Just nitpick

stefcameron: It would be better, yes. Doing it that way would help us stick to a common pixel grid, which could help with alignment. Like I noticed the `maxWidth` of the `WelcomeInner` component is 717px -- why 717px? Seems odd (pardon the pun! 😉)

Not a blocker, but so far, I've tried to do as Bogdan has indicated.

src/renderer/components/GlobalPage/WelcomeView.js
```diff
@@ -0,0 +1,145 @@
...
+      content: '"-"',
+      marginRight: 4,
+    },
+  };
+});
+
+const WelcomeLink = styled.a(function () {
+  return {
+    textDecoration: 'none',
+    color: 'var(--blue)',
```
stefcameron, line 94: Looks like Lens is using `--primary` for links. It's the same HEX value, but let's use the same variable as Lens is using.

src/renderer/components/GlobalPage/WelcomeView.js
```diff
@@ -0,0 +1,145 @@
...
+  //
+  // RENDER
+  //
+
+  return (
+    <WelcomeContainer>
+      <WelcomeWrapper>
+        <WelcomeInner>
+          <WelcomeIconWrapper>
+            <ContainerCloudIcon size={100} fill="#FFFFFF" />
```
stefcameron, line 120: This fill should use a CSS Var so that it updates to a dark shade if the theme changes to Light mode. Looks like that would be `--textColorAccent`, which is white in Dark mode, and near black in Light mode.

src/strings.ts
```diff
@@ -402,3 +402,23 @@ export const catalog: Dict = {
...
+export const welcome: Dict = {
+  title: () =>
+    'Welcome to the <br/>Mirantis Container Cloud for Lens Extention',
+  description: () =>
+    'This extension enables you to connect to multiple Mirantis Container Cloud management clusters through Lens. You can now leverage Lens for basic lifecycle management operations such as:',
+  listItems: () => `
+    <li>Synchronizing projects to view and manage their resources in the Lens Catalog</li>
+    <li>Monitoring real-time cluster status and resource utilization in Lens</li>
+    <li>Creating and deleting clusters in your projects from Lens</li>
+  `,
```
stefcameron, line 415: When the string is HTML (that would require `dangerouslySetInnerHTML` to use it), I add "Html" to the end of the string name, like `titleHtml` and `listItemsHtml` in this case. See `login.sso.messageHtml` as an example.

src/strings.ts
```diff
@@ -402,3 +402,23 @@ export const catalog: Dict = {
...
+  link: {
+    label: () => 'And more!',
+    href: () =>
+      'https://github.com/Mirantis/lens-extension-cc/blob/master/README.md',
```
stefcameron, line 419: A URL is more of a constant than a string. Could we put this in `constants.js`?

yarn.lock
```diff
@@ -1233,6 +1233,18 @@
...
+  resolved "https://registry.yarnpkg.com/@mirantis/mds-button/-/mds-button-0.5.4.tgz#da84713d4fac813d40483ac59b056f2e57024600"
+  integrity sha512-fXSxLSJccDjK/NKAix/IMY31UD9zbq1FozfoEPk/H6fUUTX7XGRL48dLL2P/BtldD2lOvXp0Nq1RmHPF4Me8mQ==
+
+"@mirantis/mds-theme@^0.10.2":
+  version "0.10.2"
+  resolved "https://registry.yarnpkg.com/@mirantis/mds-theme/-/mds-theme-0.10.2.tgz#6f443cbffe39c89085a8dbb5e81bc369ed93685d"
+  integrity sha512-eX2y3Md3+xjeW8bGlWyVYinMc+rEPCKc2eziaptD90bMull1JSv75+dZXS7X8RtuVc4nkB3rUGQe3lDOYfuulg==
+  dependencies:
+    js-cookie "^3.0.1"
+
```
stefcameron, line 1247: Unfortunately, we can't use MDS components in Lens. MDS packages are private, and this extension is public. Also, the style will be very different from what Lens components look like. We have to use Lens components only.

stefcameron (changes requested):

src/renderer/components/GlobalPage/WelcomeView.js
AntonNos: Fixed

src/renderer/components/GlobalPage/WelcomeView.js
AntonNos: Fixed

src/strings.ts
AntonNos: Fixed

src/strings.ts
AntonNos: Fixed

yarn.lock
AntonNos: Fixed

stefcameron (approved):
Thanks for making those changes!

PR #671, opened 2022-01-13
[PRODX-20596]: added synchronized block

AntonNos:
Screenshot 2022-01-13 at 14 47 34


src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,199 @@ const Title = styled.h3(() => ({
...
+            onClick={sortByName}
+            isRotated={nextSortType === 'DESC' ? true : false}
+          >
+            <CaretIcon />
+          </SortButton>
+        </ProjectsHead>
+        <ProjectsBody>
+          <ProjectsList>
+            {projectsList.map((project, index) => (
+              <li key={index}>
```
bnalisnikovskiy, line 190: I guess `project.projectName` is better for `key`

src/renderer/components/Accordion/ChevronIcon.js
```diff
@@ -0,0 +1,16 @@
...
+      viewBox="0 0 24 24"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        d="M15.7131 12L9.70209 5.98999L8.28809 7.40399L12.8881 12.004L8.28809 16.597L9.70209 18.011L15.7131 12Z"
+        fill="var(--textColorSecondary)"
+      />
+    </svg>
+  );
```
stefcameron, line 15: Icons should come from Material (where Lens gets all its icons). So this would be one of their "chevron" icons, like https://fonts.google.com/icons?selected=Material%20Icons%3Achevron_right%3A

src/renderer/components/Checkbox/CheckIcon.js
```diff
@@ -0,0 +1,21 @@
...
+      style={{
+        stroke: 'rgb(23, 34, 47)',
+        fill: 'none',
+      }}
+    >
+      <g>
+        <path d="M1 4.5L3.98039 7L9 1" strokeWidth="2"></path>
+      </g>
+    </svg>
+  );
```
stefcameron, line 20: https://fonts.google.com/icons?selected=Material%20Icons%3Acheck%3A

src/renderer/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,100 @@
+import { useState, useEffect } from 'react';
```
stefcameron, line 1: Nice job on a Checkbox component, but why do we have our own Checkbox? Lens already provides us with a Checkbox component. For example, we use one in the existing code in the `ClusterList` component: https://github.com/Mirantis/lens-extension-cc/blob/4a9beeee39c8a93213d1227190521e146e5d4326/src/renderer/components/GlobalPage/ClusterList.js#L137

Before making any components of our own, we need to check what Lens offers and use those first. The Onboarding wiki I wrote provides information about where to find what components Lens provides.

src/renderer/components/GlobalPage/CaretIcon.js
```diff
@@ -0,0 +1,16 @@
...
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        d="M12 9.5L7 14.5L17 14.5L12 9.5Z"
+        fill="var(--textColorSecondary)"
+      />
+    </svg>
+  );
+};
```
stefcameron, line 16: This is probably https://fonts.google.com/icons?selected=Material%20Icons%3Aarrow_drop_up%3A

src/renderer/components/Accordion/Accordion.js
```diff
@@ -0,0 +1,48 @@
...
+  cursor: pointer;
+  margin-right: ${layout.grid}px;
+  transform: ${({ isOpen }) => (isOpen ? 'rotate(90deg)' : 'rotate(0deg)')};
+`;
+
+const AccordionItem = styled.div`
+  overflow: hidden;
+`;
+
+export const Accordion = ({ title, children }) => {
```
stefcameron, line 24: Please make sure Lens doesn't already have this kind of component. I don't think so, but we better have a look to be sure.

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,199 @@ const Title = styled.h3(() => ({
...
+  };
+
   return (
     <Content>
       <Title>{synchronizeBlock.title()}</Title>
+      <Projects>
+        <ProjectsHead>
+          <Checkbox
+            handleChildCheckbox={handleChildCheckbox}
+            label="Project name"
```
stefcameron, line 177: Please move this string into the `strings.ts` module.

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,199 @@ const Title = styled.h3(() => ({
...
+    ],
+  },
+];
+
 export const SynchronizeBlock = () => {
+  // @type {boolean} states of all checkbox
+  const [isAllChecked, setIsAllChecked] = useState(false);
+
+  // @type {object} sorted object of projects
+  const [projectsList, setProjectsList] = useState(mockedProjects);
```
stefcameron, line 146: Ultimately, the list of projects and their data should be coming from a prop that will be passed into this component. Is that work for later, or something you forgot to do?

stefcameron (changes requested):
Nice work on the layout; looks great, visually! 😄

src/renderer/components/GlobalPage/SynchronizeBlock.js
AntonNos: I mocked data only for now to test component and check UI part. Of course I will delete mocked data once all functionality will be done:)

AntonNos:
> Nice work on the layout; looks great, visually! 😄

Thank you! Going to fix all your comments

src/renderer/components/GlobalPage/SynchronizeBlock.js
bnalisnikovskiy: You can pass it as a prop and set `mockedProjects` as default prop in `defaultPropTypes` 
But I think this structure will be changed. 
As I can see we will pass something like this 
`selectedCloud: {
cloud: Cloud
namespaces: [...],
clusters: [...],
sshKeys: [...],
...
}`

And action `cloudStore.clouds[selectedCloud.cloud.cloudUrl] = selectedCloud.cloud` will happen here after 'sync' button click (not at `ConnectionBlock`, as we do now). In ConnectionBlock we create `new Cloud()`, fetch all needed data, construct `selectedCloud` and set it to parent (`AddCloudInstance`) local state.
Then it should looks like `<SynchronizeBlock cloud={selectedCloud} />` and all work to store on disk and sync should be called from `SynchronizeBlock` 

Does it make sense @stefcameron ?

src/renderer/components/Checkbox/Checkbox.js
AntonNos: Yes, I saw that checkbox in `ClusterList`, but looks like we need to have custom icon (minus) in checkbox if `isAllCkecked` is false. Plus need to add handler to `Checkbox` component to handle state in parent component  (which I can't do with Lens Checkbox component). That why I created our own `Checkbox`

src/renderer/components/Accordion/Accordion.js
AntonNos: Okay, thank you! I will check them in the future work. Looks like we don't have `Accordion` (or similar) component in Lens

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: I think we ended-up chatting about this in Slack. We should have an `ExtendedCloud` class that is created with a `Cloud` object.

ExtendedCloud can own the sync timer, have properties like `clusterCount`, `credentialCount`, etc., and fire "update" events (similar to how Cloud fires "status update" events) whenever that data is updated. ExtendedCloud can contain all the API access logic to get all the metadata we want to display for a given Cloud so that we keep all that out of the Cloud class itself since it's already pretty complex.

Then we can simply give an ExtendedCloud object to something like the SyncBlock and it can just use its properties and subscribe to its event in order to display what it needs to. `extendedCloud.cloud.syncNamespaces` and `extendedCloud.cloud.syncAll`, for example, will tell us which projects are synced. `extendedCloud.clusterCount` will tell us the number of clusters. Etc. `<SyncBlock extCloud={new ExtendedCloud(cloud)} />`

Later, when we get to the SyncManager (step 12 in the Google Doc), we can use ExtendedCloud to get more than just counts, but all the objects.

So perhaps we'll want to end-up with a context provider (similar to ClusterDataProvider that exists today and will go away) that provides ExtendedCloud objects to the entire "app" (i.e. extension). That provider observes CloudStore.clouds for changes, and updates its list of extended cloud objects by doing something like `cloudStore.clouds.map((c) => new ExtendedCloud(c))`.

src/renderer/components/Accordion/Accordion.js
stefcameron: Thanks for checking!

src/renderer/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,100 @@
...
+  useEffect(() => {
+    setIsChecked(isAllChecked);
+  }, [isAllChecked]);
+
+  return (
+    <CheckboxItem>
+      <CheckboxControlPart>
+        <CheckboxControl
+          role="checkbox"
+          aria-checked="false"
```
stefcameron, line 74: This attribute should be one of "true", "mixed", or "false" depending on the Checkbox's state.

src/renderer/components/Checkbox/Checkbox.js
stefcameron: Ah, yes, we need a tri-state checkbox and Lens' checkbox is only dual-state. But I don't see any indication of a third state in the code. It's all binary, as far as I can tell. Is that something you're planning on adding later?

src/renderer/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,100 @@
...
+}));
+
+export const Checkbox = ({ label, handleChildCheckbox, isAllChecked }) => {
+  // @type {boolean} states of checkbox
+  const [isChecked, setIsChecked] = useState(false);
+
+  // pass checkbox state to parent component
+  // need to have handleChildCheckbox method in parent
+  useEffect(() => {
+    handleChildCheckbox && handleChildCheckbox(isChecked);
```
stefcameron, line 61: Ideally, a Checkbox shouldn't know about any other Checkbox. The fact that all checkboxes (projects) are checked, including the parent, or that some checkboxes are checked and the parent is partially checked, should only be the concern of the SynchronizeBlock component. The SynchronizeBlock has access to all checkboxes, and so should own/set/manage the state of each one.

src/renderer/components/Checkbox/Checkbox.js
stefcameron: But we aren't concerned about accessibility in Lens. So it might be better just to remove all the `aria-` attributes.

src/renderer/components/Checkbox/Checkbox.js
AntonNos: Yes, I just see that 3rd state when answered your comment. I'm working on it right now and will and it in next commit (together with all "after review" fixes)

AntonNos: Added 3rd state

src/renderer/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,102 @@
...
+}) => {
+  return (
+    <CheckboxItem>
+      <CheckboxControlPart>
+        <CheckboxControl isChecked={isCheckedFromParent}>
+          {isMinusIcon && isCheckedFromParent && (
+            <Component.Icon
+              material="remove"
+              style={{
+                color: 'rgb(23, 34, 47)',
```
stefcameron, line 69: Will this (and line 78) hard coded color work when the them is changed from dark to light or vice versa?

src/renderer/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,102 @@
...
+        </CheckboxLabel>
+      </CheckboxControlPart>
+    </CheckboxItem>
+  );
+};
+
+Checkbox.propTypes = {
+  label: PropTypes.string,
+  onChange: PropTypes.func,
+  isCheckedFromParent: PropTypes.bool,
```
stefcameron, line 100: What does this prop mean? Can we just call it "checked"? Having this "parent" reference in the name seems to make this more complicated than it is: A tri-state checkbox.

src/renderer/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,102 @@
...
+  isMinusIcon: PropTypes.bool,
```
stefcameron, line 101: So if `isMinusIcon` is true, then it's indeterminate regardless of `isCheckedFromParent`? `isMinusIcon` seems too implementation-specific -- like what if we changed it to be a filled square instead of a hyphen when the component is in an indeterminate state...?

How about we merge `isMinusIcon` and `isCheckedFromParent` into one prop, `value`, and provide an export that is an enumeration of supported values "CHECKED", "UNCHECKED", "MIXED"?

```javascript
export const values = {
  CHECKED: "CHECKED",
  UNCHECKED: "UNCHECKED",
  MIXED: "MIXED"
};
```

Now we could have `<Checkbox value={values.MIXED} />` to set it into its indeterminate state, regardless of what that looks like visually. And in the code, there's no confusion as to how the Checkbox should render: Checkmark, empty, or hyphen.

And finally, since Lens already has a `<Checkbox>` component, we should really name this one something different so there's no confusion. Maybe "TriStateCheckbox" or something?

src/renderer/components/Checkbox/Checkbox.js
AntonNos: Fixed

src/renderer/components/Checkbox/Checkbox.js
AntonNos: Fixed

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,354 @@ const Title = styled.h3(() => ({
...
+
+  if (!projectsList) {
+    return null;
+  }
+
+  const values = {
+    CHECKED: 'CHECKED',
+    UNCHECKED: 'UNCHECKED',
+    MIXED: 'MIXED',
+  };
```
bnalisnikovskiy, line 229: Move away from component. And 'values' isn't very proper name. eg `PARENT_VALUES = Object.freeze({...})`

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,354 @@ const Title = styled.h3(() => ({
...
+    if (childrenCheckboxes.some((el) => el === false)) {
+      return true;
+    }
+    return currentCheckboxState.parent;
+  };
+
+  const onChangeHandler = (index) => {
+    if (isNaN(index)) {
+      setCurrentCheckboxState({
+        ...currentCheckboxState,
```
bnalisnikovskiy, line 262: this is not necessary. You are rewrite all values anyway

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,354 @@ const Title = styled.h3(() => ({
...
+        children: currentCheckboxState.children.map(
+          (el) => (el = !currentCheckboxState.parent)
+        ),
+      });
+    } else {
+      const modifiedChildrenCheckboxes = currentCheckboxState.children.map(
+        (el, i) => (index === i ? (el = !el) : el)
+      );
+      setCurrentCheckboxState({
+        ...currentCheckboxState,
```
bnalisnikovskiy, line 273: this is not necessary. You are rewrite all values anyway

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,354 @@ const Title = styled.h3(() => ({
...
+        parent: !currentCheckboxState.parent,
+        children: currentCheckboxState.children.map(
+          (el) => (el = !currentCheckboxState.parent)
+        ),
+      });
+    } else {
+      const modifiedChildrenCheckboxes = currentCheckboxState.children.map(
+        (el, i) => (index === i ? (el = !el) : el)
```
bnalisnikovskiy, line 270: It may be simpler ` (el, i) => (index === i ? !el : el)`

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,354 @@ const Title = styled.h3(() => ({
...
+    return currentCheckboxState.parent;
+  };
+
+  const onChangeHandler = (index) => {
+    if (isNaN(index)) {
+      setCurrentCheckboxState({
+        ...currentCheckboxState,
+        parent: !currentCheckboxState.parent,
+        children: currentCheckboxState.children.map(
+          (el) => (el = !currentCheckboxState.parent)
```
bnalisnikovskiy, line 265: It may be simpler `() => !currentCheckboxState.parent`

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,354 @@ const Title = styled.h3(() => ({
...
+    if (
+      currentCheckboxState.parent &&
+      currentCheckboxState.children.some((el) => el === false) &&
+      currentCheckboxState.children.some((el) => el === true)
+    ) {
+      return values.MIXED;
+    }
+    if (currentCheckboxState.parent) {
+      return values.CHECKED;
+    } else {
```
bnalisnikovskiy, line 290: `else` is not necessary. Just return

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +22,354 @@ const Title = styled.h3(() => ({
...
+                color: 'var(--textColorSecondary)',
+                fontSize: 'calc(var(--font-size) * 1.8)',
+                marginTop: layout.grid / 2,
+              }}
+            />
+          </SortButton>
+        </ProjectsHead>
+        <ProjectsBody>
+          <ProjectsList>
+            {projectsList.map((namespace, index) => (
```
bnalisnikovskiy, line 328: it would be easier if instead of indexes we will use namespace names. In the result we need to know their names anyway.
So instead of `[true, false]` array we would have `{ name1: true, name-two: false }`.
It also may be more safe since we have sorting (change order) here

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+              style={{
+                color: 'var(--inputControlBackground)',
+                fontSize: 'calc(var(--font-size) * 1.14)',
+              }}
+            />
+          )}
+          {value === 'CHECKED' && (
+            <Component.Icon
+              material="check"
+              style={{
```
bnalisnikovskiy, line 72: Lets move styles Objects from Component (here and on line 63)
Just call them `removeIconStyle`, `checkIconStyle` etc, and keep outside of component

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+                fontSize: 'calc(var(--font-size) * 1.14)',
+              }}
+            />
+          )}
+        </CheckboxControl>
+        <CheckboxLabel>
+          <CheckboxField
+            type="checkbox"
+            checked={value !== 'UNCHECKED' ? true : false}
+            onChange={() => onChange()}
```
bnalisnikovskiy, line 83: You can use just `onChange`. Arrow func isn't needed here, as I can see

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+  );
+};
+
+TriStateCheckbox.propTypes = {
+  label: PropTypes.string,
+  onChange: PropTypes.func,
+  isChecked: PropTypes.bool,
+  isMinusIcon: PropTypes.bool,
+  value: PropTypes.string,
+};
```
bnalisnikovskiy, line 98: If propTypes aren't required you have to set defaultProps. 
So `.isRequired` or defaultProp

src/renderer/components/Accordion/Accordion.js
```diff
@@ -0,0 +1,56 @@
...
+    <div>
+      <AccordionTitle isOpen={isOpen}>
+        <AccordionButton
+          type="button"
+          isOpen={isOpen}
+          onClick={() => setIsOpen(!isOpen)}
+        >
+          <Component.Icon
+            material="chevron_right"
+            style={{
```
bnalisnikovskiy, line 40: Better to keep style objects outside of component to avoid extra creations

bnalisnikovskiy (changes requested):

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: I meant for this enumeration to be defined by, and exported from, TriStateCheckbox, since that's the component that uses these values. The enumeration should be _imported_ here, and then used throughout the code like `checkValues.CHECKED` to check one of the TriStateCheckbox components.

The values are not specific to the parent. If `values` alone is too generic, then maybe `checkValues`?

src/renderer/components/GlobalPage/AddCloudInstance.js
```diff
@@ -5,6 +5,8 @@ import { ConnectionBlock } from './ConnectionBlock';
 import { SynchronizeBlock } from './SynchronizeBlock';
 import { CloseButton } from '../CloseButton/CloseButton';
 
+// import { mockedProjects } from './mockedProjects';
```
stefcameron, line 8: This comment should be removed.

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+  overflowWrap: 'break-word',
+  hyphens: 'auto',
+  overflow: 'hidden',
+}));
+
+export const TriStateCheckbox = ({ label, onChange, value }) => {
+  return (
+    <CheckboxItem>
+      <CheckboxControlPart>
+        <CheckboxControl isChecked={value !== 'UNCHECKED'}>
```
stefcameron, line 59: "UNCHECKED" should come from the enumeration, like `checkValues.UNCHECKED`, instead of being hard coded.

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+          {value === 'MIXED' && (
```
stefcameron, line 60: Same for "MIXED"

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+            <Component.Icon
+              material="remove"
+              style={{
+                color: 'var(--inputControlBackground)',
+                fontSize: 'calc(var(--font-size) * 1.14)',
+              }}
+            />
+          )}
+          {value === 'CHECKED' && (
```
stefcameron, line 69: Same for "CHECKED"

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+                color: 'var(--inputControlBackground)',
+                fontSize: 'calc(var(--font-size) * 1.14)',
+              }}
+            />
+          )}
+        </CheckboxControl>
+        <CheckboxLabel>
+          <CheckboxField
+            type="checkbox"
+            checked={value !== 'UNCHECKED'}
```
stefcameron, line 82: Same for "UNCHECKED"

src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -0,0 +1,98 @@
...
+      </CheckboxControlPart>
+    </CheckboxItem>
+  );
+};
+
+TriStateCheckbox.propTypes = {
+  label: PropTypes.string,
+  onChange: PropTypes.func,
+  isChecked: PropTypes.bool,
+  isMinusIcon: PropTypes.bool,
```
stefcameron, line 96: Now that we have `value` prop, we don't need `isChecked` and `isMinusIcon`.

stefcameron (commented):
TriStateCheckbox is looking good...!

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +25,364 @@ const Title = styled.h3(() => ({
...
+};
+
+const checkboxesStateObj = (extCloud) => {
+  return extCloud.namespaces.reduce((acc, namespace) => {
+    acc[namespace.name] = false;
+    return acc;
+  }, {});
+};
+
+export const SynchronizeBlock = ({ extCloud }) => {
```
bnalisnikovskiy, line 218: You should check  if `extCloud` is present and return null if it's not

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +25,364 @@ const Title = styled.h3(() => ({
...
+  // @type {string} sort by name order
+  const [nextSortType, setNextSortType] = useState('');
+
+  // @type {object} checkboxes state
+  const [сheckboxesState, setCheckboxesState] = useState({
+    parent: false,
+    children: checkboxesStateObj(extCloud),
+  });
+
+  if (!projectsList) {
```
bnalisnikovskiy, line 231: See my comment above. Instead of this part you can add 
`if(extCloud?.namespaces) {
return null
}`

src/renderer/components/GlobalPage/SynchronizeBlock.js
AntonNos: We will have linter error if we add `if` statement before `useState`

src/renderer/components/GlobalPage/SynchronizeBlock.js
AntonNos: We will have linter error if we add `if` statement before `useState`

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: That's true. React "rules of hooks" means we can't conditionally execute a hook. You'll also have to deal with the fact that the ExtendedCloud object may not yet be loaded when you get it. You'll need to check its `loading` property and subscribe to its LOADING_CHANGE event. But not just on first load of SynchronizeBlock. This should be done such that at any time, even after the SynchronizeBlock has been rendered, the ExtendedCloud object might go into loading state (because it'll be refreshing its data every 5 minutes).

@bnalisnikovskiy This reminds me, ExtendedCloud should have another event like DATA_UPDATED to signal when its data (i.e. `namespaces` property) has been updated.

The point is that the `loading` property (and LOADING_CHANGE event) represent the initial data load where any UI rendering the ExtendedCloud __must__ show a spinner because there's no data. But once we have data, 5 minutes later, we'd going to refresh it -- but we'll still have data to show while we're refreshing (the known/existing data). So UI that's rendering the ExtendedCloud instance will want to continue showing the current data right up until there's new data available. Otherwise, every 5 minutes, the whole UI would go into "spinner mode", which won't be nice. So components rendering an ExtendedCloud object will need to be notified when new data is available (i.e. ExtendedCloud.namespaces has been updated). To do that, they can listen for the DATA_UPDATED event.

And probably if there's an error during refresh, we should log the error and just not do anything (i.e. don't fire the event, and don't touch the existing `ExtendedCloud#namespaces` property so that components rendering it will continue to render the last-known data. Or we could set an `error` property with a message and set `namespaces = undefined`, but that would further complicate any component that renders an ExtendedCloud object.

@AntonNos The point I'm making is that if `extCloud.loading === true`, SynchronizeBlock will need to show a spinner (because `extCloud.namespaces` will be undefined) and listen for the LOADING_CHANGE event. Once the data is loaded, it'll also need to listen for DATA_UPDATED and then update according to the new `extCloud.namespaces` object.

src/strings.ts
```diff
@@ -121,6 +121,13 @@ export const connectionBlock: Dict = {
 };
 export const synchronizeBlock = {
   title: () => 'Select projects to synchronize',
+  synchronizeButtonLabel: () => 'Synchronize selected projects',
+  checkAllCheckboxLabel: () => 'Project name',
+  checkboxesDropdownLabels: {
+    clusters: () => 'clusters',
+    sshKeys: () => 'ssh keys',
```
stefcameron, line 128: Let's make "ssh" -> "SSH" (all caps) since it's an acronym.

stefcameron (commented):
TriStateCheckbox is looking solid now, thank you!

src/renderer/components/GlobalPage/SynchronizeBlock.js
bnalisnikovskiy: I think it's a bit overcomplicated. I thought do add check on AddCloudInstance level, and don't show SynchronizeBlock until ExtCloud will be loaded.

stefcameron: We'll still need this mechanism in ExtendedCloud for the SyncView. If we're really only going to use the SynchronizeBlock for the sole purpose of the AddCloudInstance wizard, then OK, I agree it only needs to deal with the `loading` property. But it seems to me we'll want to use it in the SyncView in "read-only" mode (without the checkboxes) since we need to show the exact same UI there (just without the checkboxes) with collapsing section to show details under projects, etc. So then we'll need it to handle what I just described...

bnalisnikovskiy: You're right. Noted about DATA_UPDATED. I'll add that (if not find better solution) on next step, together with `ExtendedCloudProvider`

bnalisnikovskiy (approved):

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: Will the handling of the LOADING_CHANGE event be added in a subsequent PR?

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -14,10 +25,370 @@ const Title = styled.h3(() => ({
...
+      clustersCount: 0,
+      sshKeys: [],
+      sshKeysCount: 0,
+      credentials: {
+        awscredential: [],
+        allCredentialsCount: 1,
+        byocredential: [],
+        openstackcredential: [{}],
+      },
+    },
```
stefcameron, line 119: I would feel more comfortable committing this to master (and so "forever" in the history of this public repo) if the data was at least anonymized. Replace actual email addresses and URLs and security tokens with fake strings.

Better yet, I'd prefer mock data be located in a module in https://github.com/Mirantis/lens-extension-cc/tree/master/test/mocks and imported instead of being set in a component like this.

stefcameron (changes requested):

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: This is technically a security breach as it is, since this PR is publicly posted on the repo for the entire world to see, and includes tokens that were actually valid at one point in time.

AntonNos: Fixed

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: Confirmed this will be handled in a future PR.

stefcameron (approved):
Good to merge! Thank you for your work on this.

PR #672, opened 2022-01-13
WIP added onCheckStatusChange

bnalisnikovskiy:
I decided  create separate branch regarding to this comment. Just to not swamp original PR, there are enough other comments)
 
image

Sorry, I tried, but couldn't got a full picture how it should work. Hence, without understanding I can't made that changes.  Instead, I've created a separate PR with another possible implementation. 

Probably, my brain get used to work in another way. If you implemented your version there, I would glad to see it.

### DIFF
--- a/src/common/Cloud.js
+++ b/src/common/Cloud.js
@@ -139,8 +139,13 @@ export class Cloud {
         return _connectError;
       },
       set(newValue) {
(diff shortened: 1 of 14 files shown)


stefcameron:
See https://github.com/Mirantis/lens-extension-cc/pull/673

PR #673, opened 2022-01-13
Add status change event dispatching to Cloud

stefcameron:
I also fixed a few other issues I found, including how some of
the new Cloud properties were not setting the `_changed` flag.
I did that for consistency with the rest of the code, but I don't
think we need that property anymore, so we could clean that up.

It's only used in one place elsewhere in the code, and that place
could just assume it's always changed.


src/common/Cloud.js
```diff
@@ -162,12 +293,24 @@ export class Cloud {
...
+      },
+      set(newValue) {
+        DEV_ENV &&
+          rtv.verify(
+            { token: newValue },
+            { token: Cloud.specTs.syncNamespaces }
+          );
+        if (_syncNamespaces !== newValue) {
+          _syncNamespaces = newValue;
+          _changed = true;
```
bnalisnikovskiy, line 307: With this structure we can't use `.push` and should always reassign new array. Is it okay? 
That's a reason why I added it as a `value: []`. It gives more flexibility

src/common/Cloud.js
```diff
@@ -571,27 +716,30 @@ export class Cloud {
...
         }
+
+        this.connecting = false;
+        this.dispatchEvent(CLOUD_EVENTS.STATUS_CHANGE, this);
       }.bind(this);
 
       addExtEventHandler(EXT_EVENT_OAUTH_CODE, handler, state);
     } else {
       this.connecting = false;
+      this.dispatchEvent(CLOUD_EVENTS.STATUS_CHANGE, this);
```
bnalisnikovskiy, line 735: As I can see all `this.dispatchEvent()` in `connect` are redundant.  We call `dispatchEvent` in `connecting` setter and all dispatchEvents placed right next to `connecting`. I understand that is okay ,no harm, because of EventListener construction, but no benefits here as well

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -57,20 +61,23 @@ export const ConnectionBlock = () => {
...
+      } else {
         setLoading(false);
+        newCloud.removeEventListener(
+          CLOUD_EVENTS.STATUS_CHANGE,
+          statusListener
+        );
+        if (newCloud.status === CONNECTION_STATUSES.CONNECTED) {
+          cloudStore.clouds[normUrl] = newCloud;
+        } else {
+          checkError(newCloud);
```
bnalisnikovskiy, line 76: Nice

src/common/Cloud.js
```diff
@@ -116,10 +126,109 @@ export class Cloud {
...
+       * @method dispatchEvent
+       * @param {string} name Event name.
+       * @param {Array} [params] Parameters to pass to each handler, if any.
+       */
+      dispatchEvent: {
+        enumerable: false,
+        value(name, ...params) {
+          const event = _eventQueue.find((e) => e.name === name);
+          if (event) {
+            event.params = params;
```
bnalisnikovskiy, line 215: Aha, so we update params (`this`) each time. Nice

bnalisnikovskiy (approved):

bnalisnikovskiy:
Not tested yet, but code looks good. Left a one question and one nitpick (not so important)

src/common/Cloud.js
stefcameron: We can still use `.push` with this, but I see what you mean. It leaves the door open to set a whole new array, while using `value: []` doesn't. I'll change this back.

src/common/Cloud.js
stefcameron: This is true. Technically, we shouldn't need any explicit dispatch calls here because all the properties that get modified will also dispatch the same event. I thought being more explicit about it was better, but maybe not.

stefcameron:
> Not tested yet, but code looks good. Left a one question and one nitpick (not so important)

Thanks for having a look! I updated after your comments.

src/common/Cloud.js
```diff
@@ -293,25 +292,11 @@ export class Cloud {
...
-          rtv.verify(
-            { token: newValue },
-            { token: Cloud.specTs.syncNamespaces }
-          );
-        if (_syncNamespaces !== newValue) {
-          _syncNamespaces = newValue;
-          _changed = true;
-        }
-      },
+      value: [],
```
bnalisnikovskiy, line 281: I think in this case we also need set `writable: true`. It allows us to use all operations, like `cloud.syncNamespaces = cloud.syncNamespaces.filter(c => c !== name)`, etc

stefcameron: Right, well, in that case, we should go back to the getter/setter like all other properties. Once we make it writable, there's no difference in what you can do from an external perspective (modify the array, or replace it altogether).

src/common/Cloud.js
```diff
@@ -278,7 +279,19 @@ export class Cloud {
...
+        return _syncNamespaces;
+      },
+      set(newValue) {
+        DEV_ENV &&
+          rtv.verify(
+            { token: newValue },
+            { token: Cloud.specTs.syncNamespaces }
+          );
+        if (_syncNamespaces !== newValue) {
+          _syncNamespaces = newValue;
```
bnalisnikovskiy, line 292: yeah, setter better. instead of `push` we always can do `cloud.syncNamespaces = [...cloud.syncNamespaces, 'newName']`

PR #676, opened 2022-01-18
PRODS-20985 Create ExtendedCloud

bnalisnikovskiy:
This is a WIP version. Recommendations are welcomed

### DIFF
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -269,6 +269,7 @@ export class ExtendedCloud {
   constructor(cloud) {
     const _eventListeners = {}; // map of event name to array of functions that are handlers
     const _eventQueue = []; // list of `{ name: string, params: Array }` objects to dispatch
+    let _error = null;
 
     // asynchronously dispatches queued events on the next frame
     const _scheduleDispatch = () => {
@@ -401,6 +402,18 @@ export class ExtendedCloud {
         _namespaces = newValue;
       },
     });
+
+    Object.defineProperty(this, 'error', {
+      enumerable: true,
+      get() {
+        return _error;
+      },
+      set(newValue) {
+        _error = newValue;
+      },
+    });
+
+    setTimeout(() => this.init());
   }
 
   startUpdateCloudByTimeOut() {
@@ -422,6 +435,8 @@ export class ExtendedCloud {
    */
   async init(loadAll) {
     this.loading = true;
+    this.error = null;
+
     const { error: nameSpaceError, namespaces } = await _fetchNamespaces(
       this.cloud,
       loadAll
@@ -432,7 +447,8 @@ export class ExtendedCloud {
         `_fetchNamespaces error: ${nameSpaceError.message}`,
         nameSpaceError
       );
-      throw new Error(nameSpaceError?.message);
+      this.error = nameSpaceError.message;
+      this.loading = false;
     } else {
       const { error: clustersError, clusters } = await _fetchClusters(
         this.cloud,
@@ -442,27 +458,28 @@ export class ExtendedCloud {
         this.cloud,
         namespaces
       );
-      const { credentials, error: credentialsError } = await _fetchCredentials(
+      const { error: credentialsError, credentials } = await _fetchCredentials(
         this.cloud,
         namespaces
       );
 
-      let error = clustersError || sshKeysError || credentialsError;
+      const error = clustersError || sshKeysError || credentialsError;
       if (error) {
-        this.loading = false;
         logger.error(
           'extendedCloud.init()',
           `Fetched data contains an error: ${error.message}`,
           error
         );
-        throw new Error(error.message);
+        this.error = error.message;
+        this.loading = false;
       } else {
-        const nameSpaces = namespaces.map((name) => {
+        this.namespaces = namespaces.map((name) => {
           const nameSpaceClusters = clusters.filter(
             (c) => c.namespace === name
           );
           const nameSpaceSshKeys =
             sshKeys.find((key) => key.namespace === name)?.items || [];
+
           return {
             name,
             clusters: nameSpaceClusters,
@@ -473,7 +490,6 @@ export class ExtendedCloud {
           };
         });
 
-        this.namespaces = nameSpaces;
         this.loading = false;
         return this;
       }
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -1,0 +1,357 @@
+import * as rtv from 'rtvjs';
+import { Cloud } from './Cloud';
+import { filter, find, flatten } from 'lodash';
+import { authedRequest, extractJwtPayload } from '../renderer/auth/authUtil';
+import * as strings from '../strings';
+import { Namespace } from '../renderer/store/Namespace';
+import { logger } from '../util/logger';
+import { Cluster } from '../renderer/store/Cluster';
+
+const credentialTypes = [
+  'awscredential',
+  'byocredential',
+  'openstackcredential',
+];
+
+const _deserializeCredentialsList = function (body) {
+  if (!body || !Array.isArray(body.items)) {
+    return {
+      error: 'Failed to parse Credentials payload: Unexpected data format.',
+    }; // TODO move to string
+  }
+
+  return body.items;
+};
+
+const _fetchCredentials = async function (cloud, namespaces) {
+  const results = await Promise.all(
+    credentialTypes.map(async (entity) => {
+      return await Promise.all(
+        namespaces.map(async (namespaceName) => {
+          const { body, error } = await authedRequest({
+            baseUrl: cloud.cloudUrl,
+            cloud,
+            config: cloud.config,
+            method: 'list',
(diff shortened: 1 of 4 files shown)


src/common/ExtendedCloud.js
```diff
@@ -0,0 +1,479 @@
...
+   * @return {Promise<ExtendedCloud>}
+   */
+  async init(loadAll) {
+    this.loading = true;
+    const { error: nameSpaceError, namespaces } = await _fetchNamespaces(
+      this.cloud,
+      loadAll
+    );
+    if (nameSpaceError) {
+      logger.error(nameSpaceError?.error?.message);
```
stefcameron, line 434: Missing context parameter here.

src/common/ExtendedCloud.js
```diff
@@ -0,0 +1,479 @@
...
+      );
+      const { credentials, error: credentialsError } = await _fetchCredentials(
+        this.cloud,
+        namespaces
+      );
+
+      let error = clustersError || sshKeysError || credentialsError;
+      if (error) {
+        this.loading = false;
+        logger.error(error.message);
```
stefcameron, line 453: Missing context parameter here

src/common/ExtendedCloud.js
```diff
@@ -0,0 +1,479 @@
...
+        this.loading = false;
```
stefcameron, line 454: Already done on line 452

src/common/ExtendedCloud.js
```diff
@@ -0,0 +1,479 @@
...
+  }
+
+  return {
+    data: body.items
+      .map((item, idx) => {
+        try {
+          return new Namespace(item);
+        } catch (err) {
+          logger.error(
+            'store/ClusterDataProvider._deserializeNamespacesList()',
```
stefcameron, line 167: The "context" parameter to `logger.error()` is supposed to provide an indication as to where this log message is coming from to help with debugging. "store/ClusterDataProvider" doesn't see applicable anymore... 🙂

src/common/ExtendedCloud.js
```diff
@@ -0,0 +1,479 @@
...
+  }
+
+  return {
+    data: body.items
+      .map((item, idx) => {
+        try {
+          return new Cluster(item);
+        } catch (err) {
+          logger.error(
+            'ClusterDataProvider._deserializeClustersList()',
```
stefcameron, line 135: Same comment as line 167

src/common/ExtendedCloud.js
```diff
@@ -0,0 +1,479 @@
...
+import { authedRequest, extractJwtPayload } from '../renderer/auth/authUtil';
+import * as strings from '../strings';
+import { Namespace } from '../renderer/store/Namespace';
+import { logger } from '../util/logger';
+import { Cluster } from '../renderer/store/Cluster';
+
+export const EXTENDED_CLOUD_EVENTS = Object.freeze({
+  /**
+   * The Cloud object's connection status has changed.
+   * @param {Cloud} cloud The Cloud object.
```
stefcameron, line 13: I think this comment needs updating a bit, right? And we're passing the ExtendedCloud instance, not the Cloud.

src/renderer/components/GlobalPage/AddCloudInstance.js
```diff
@@ -36,11 +46,57 @@ const MainColumn = styled.div(function () {
...
+    };
+
+    extCl.addEventListener(
+      EXTENDED_CLOUD_EVENTS.LOADING_CHANGE,
+      loadingListener
+    );
+
+    try {
+      const result = await extCl.init(true);
+      setExtCloud(result);
```
stefcameron, line 74: This is confusing:
- We have `extCl` which is an instance of ExtendedCloud and only exists within `makeExtCloud()`.
- We have `extCloud` which is _not_ an instance of ExtendedCloud (in spite of its name _strongly_ suggesting that it is).
- We're throwing away the temporary ExtendedCloud object in favor of the `namespaces` tree that it returns, if successful.
- It's a bit strange to be monitoring a promise (from `extCl.init()`) __and__ listening for the LOADING_CHANGE event.

The idea was that we'd have an app-side context provider that watches CloudStore and sets its state to be an array of ExtendedCloud objects for every Cloud that CloudStore has.

The rest of the UI uses the context provider (similar to ClusterDataProvider) to "get" the ExtendedCloud objects and "binds" to them.

We can either immediately initiate the initial load after the ExtendedCloudProvider creates the ExtendedCloud object from a Cloud object, or some other piece of the UI can do that. For now, probably easiest to initiate the load right away.

So you'd have ExtendedCloudProvider which _observes_ CloudStore.clouds (since it's observable via Mobx). Whenver that array changes, ExtendedCloudProvider's state changes to provide an array of ExtendedCloud objects. Without doing any optimizations, it would look something like this:

```javascript
cloudStoreInstance.clouds.map((cloud) => {
  const extCloud = new ExtendedCloud(cloud);
  extCloud.load();
  return extCloud;
});
```

That resulting array becomes the `state` that's returned here: https://github.com/Mirantis/lens-extension-cc/blob/4a9beeee39c8a93213d1227190521e146e5d4326/src/renderer/store/ClusterDataProvider.js#L286

So we'd have

```javascript
  const {
    state: {
      data: { extendedClouds }, // <- we get all the extended clouds here
    },
    actions: extendedCloudActions, // we may not have any actions; not sure
  } = useExtendedClouds();
```

And for any extendedClouds item, any component anywhere can check its `loading` property, and also listen for the LOADING_CHANGE event.

stefcameron (commented):
Good direction! Some thoughts for you! 😄

src/renderer/components/GlobalPage/AddCloudInstance.js
bnalisnikovskiy: You are absolutely right about `ExtendedCloudProvider`, it will be added soon for our main SyncView page (list of all clouds). 
But in current place we don't need it. This cloud isn't stored in CloudStore yet, it is a Cloud which potentialy could be stored there after selecting namespaces at `SynchronizeBlock`. Until then we store it in component local state and we need to grab all namespaces and related data just to display correct info at `SynchronizeBlock` checkboxes

Probably I made code a bit complex in part with `extCl`, `extCloud`. Now It's updated and I left a few inline comments
There steps which look complicated, but I don't see how to make it easier or skip some of them
- We need Cloud, which comes from `ConnectionBlock` (in last PR we store it at CloudStore, but we shouldn't)
- We take that Cloud and make extCl = `new ExtendedCloud(Cloud)`. This cloud doesn't have namespaces. We cannot make API calls from constructor. 
- So we need to do `extCl.init()` which fetch and prepare all needed data. `init()` returns `this`, but we can use just `extCl`,  it is equal `this`. It's valid only if `init()` finished successfully 
- Then we store `extCl` in local state and finally we have data to show at `SynchronizeBlock`
- In `SynchronizeBlock` user selects namespaces and clicks 'sync' button => we store extCl.cloud in CloudStore => redirect to SyncView => `ExtendedCloudProvider` does all works. 
Otherwise (click close button), we lose local state and redirected to SyncView page

stefcameron: > But in current place we don't need it. This cloud isn't stored in CloudStore yet, it is a Cloud which potentialy could be stored there after selecting namespaces at `SynchronizeBlock`. Until then we store it in component local state and we need to grab all namespaces and related data just to display correct info at `SynchronizeBlock` checkboxes

Oh, right. OK, makes sense.

> * We take that Cloud and make extCl = `new ExtendedCloud(Cloud)`. This cloud doesn't have namespaces. We cannot make API calls from constructor.
> * So we need to do `extCl.init()` which fetch and prepare all needed data. `init()` returns `this`, but we can use just `extCl`,  it is equal `this`. It's valid only if `init()` finished successfully

What if you set a timeout for the next frame (i.e. `setTimeout(init)`) at the end of the constructor so that the ExtendedCloud instance essentially "auto-loads"? Then component code looks at its `loading` property and listens for the `LOADING_CHANGE` event.

bnalisnikovskiy: Sounds interesting. I'll check. Google results say that I cannot call API functions from constructor, but will see :)

stefcameron: It's probably just not good design to call API functions from a class constructor. But my proposal means they won't technically be called from the constructor. The call will be postponed to the next execution frame, which means the constructor will have finished and the class will have been instantiated, and you'll have an ExtendedCloud instance in context long before the API calls start. And they won't be executed within the context of the constructor either. Just make sure you bind `this`: `setTimeout(() => this.init())` or `setTimeout(this.init.bind(this))`

bnalisnikovskiy: Yeah, I understand that. And it works pretty well. Code is updated and ready to the final review (I hope)

stefcameron (approved):
This is looking pretty good now! Well-positioned to get into the ExtendedCloudProvider, and also add the DATA_UPDATED event we discussed on #671.

PR #683, opened 2022-01-21
PRODX-21180 Added ExtendedCloudProvider.js

bnalisnikovskiy:
WIP version ExtendedCloudProvider
 - load cloudStore.clouds - check them and create extendedClouds
 - update extendedClouds when cloudStore.clouds are changed (comparison based on `token`)
 - Update ExternalCloud additional data (namespaces, etc) each 5 min
 - Listen changes in ExternalCloud.namespaces

### DIFF
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -20,6 +20,16 @@ const credentialTypes = [
   'openstackcredential',
 ];
 
+const getErrorMessage = (error) => {
+  if (!error) {
+    return null;
+  }
+  if (typeof error === 'string') {
+    return error;
+  }
+  return error.message;
+};
+
 const _deserializeCredentialsList = function (body) {
   if (!body || !Array.isArray(body.items)) {
     return {
@@ -444,12 +454,12 @@ export class ExtendedCloud {
       loadAll
     );
     if (nameSpaceError) {
+      this.error = getErrorMessage(nameSpaceError);
       logger.error(
         'extendedCloud.init()',
-        `_fetchNamespaces error: ${nameSpaceError.message}`,
+        `_fetchNamespaces error: ${this.error}`,
         nameSpaceError
       );
-      this.error = nameSpaceError.message;
       this.loading = false;
     } else {
       const { error: clustersError, clusters } = await _fetchClusters(
@@ -467,12 +477,12 @@ export class ExtendedCloud {
 
       const error = clustersError || sshKeysError || credentialsError;
       if (error) {
+        this.error = getErrorMessage(error);
         logger.error(
           'extendedCloud.init()',
-          `Fetched data contains an error: ${error.message}`,
+          `Fetched data contains an error: ${this.error}`,
           error
         );
-        this.error = error.message;
         this.loading = false;
       } else {
         this.namespaces = namespaces.map((name) => {
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -394,7 +394,6 @@ export class ExtendedCloud {
       },
       set(newValue) {
         _cloud = newValue;
-        this.dispatchEvent(EXTENDED_CLOUD_EVENTS.DATA_UPDATED, this);
       },
     });
     Object.defineProperty(this, 'loading', {
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -125,9 +125,10 @@ const _fetchSshKeys = async function (cloud, namespaces) {
 /**
  * Deserialize the raw list of cluster data from the API into Cluster objects.
  * @param {Object} body Data response from /list/cluster API.
+ * @param {Cloud} cloud The Cloud object used to access the clusters.
  * @returns {Array<Cluster>} Array of Cluster objects.
  */
-const _deserializeClustersList = function (body) {
+const _deserializeClustersList = function (body, cloud) {
   if (!body || !Array.isArray(body.items)) {
     return { error: strings.clusterDataProvider.error.invalidClusterPayload() };
   }
@@ -136,7 +137,7 @@ const _deserializeClustersList = function (body) {
     data: body.items
       .map((item, idx) => {
         try {
-          return new Cluster(item);
+          return new Cluster(item, cloud.username);
         } catch (err) {
           logger.error(
             'ExtendedCloud._deserializeClustersList()',
@@ -262,7 +263,7 @@ const _fetchClusters = async function (cloud, namespaces) {
   let dsError;
 
   results.every((result) => {
-    const { data, error } = _deserializeClustersList(result.body);
+    const { data, error } = _deserializeClustersList(result.body, cloud);
     if (error) {
       dsError = { error };
       return false; // break
@@ -384,11 +385,16 @@ export class ExtendedCloud {
 
     let _loading = false;
     let _namespaces = null;
+    let _cloud = cloud;
 
     Object.defineProperty(this, 'cloud', {
       enumerable: true,
       get() {
-        return cloud;
+        return _cloud;
+      },
+      set(newValue) {
+          _cloud = newValue;
+          this.dispatchEvent(EXTENDED_CLOUD_EVENTS.DATA_UPDATED, this);
       },
     });
     Object.defineProperty(this, 'loading', {
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -9,6 +9,7 @@ import { Cluster } from '../renderer/store/Cluster';
 
 export const EXTENDED_CLOUD_EVENTS = Object.freeze({
   LOADING_CHANGE: 'loadingChange',
+  DATA_UPDATED: 'dataUpdated',
 });
 
 const FIVE_MIN = 300 * 1000;
@@ -207,7 +208,7 @@ const _fetchNamespaces = async function (cloud, loadAll) {
     userRoles.includes(`m:kaas:${name}@reader`) ||
     userRoles.includes(`m:kaas:${name}@writer`);
 
-  const ignoredNamespaces = cloud.config.ignoredNamespaces || [];
+  const ignoredNamespaces = cloud?.config?.ignoredNamespaces || [];
   const namespaces = filter(
     data,
     (ns) => !ignoredNamespaces.includes(ns.name) && hasReadPermissions(ns.name)
@@ -400,6 +401,7 @@ export class ExtendedCloud {
       },
       set(newValue) {
         _namespaces = newValue;
+        this.dispatchEvent(EXTENDED_CLOUD_EVENTS.DATA_UPDATED, this);
       },
     });
 
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -393,8 +393,8 @@ export class ExtendedCloud {
         return _cloud;
       },
       set(newValue) {
-          _cloud = newValue;
-          this.dispatchEvent(EXTENDED_CLOUD_EVENTS.DATA_UPDATED, this);
+        _cloud = newValue;
+        this.dispatchEvent(EXTENDED_CLOUD_EVENTS.DATA_UPDATED, this);
       },
     });
     Object.defineProperty(this, 'loading', {
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -424,7 +424,9 @@ export class ExtendedCloud {
       },
     });
 
-    setTimeout(() => this.init());
+    if (cloud.isConnected()) {
+      setTimeout(() => this.init());
+    }
   }
 
   startUpdateCloudByTimeOut() {
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -186,13 +186,12 @@ const _deserializeNamespacesList = function (body) {
  * [ASYNC] Get all existing namespaces from the management cluster.
  * @param {Cloud} cloud An Cloud object. Tokens will be updated/cleared
  *  if necessary.
- * @param {boolean} [loadAll] if true all namespaces will be loaded. Doesn't matter if we have something in syncNamespaces
  * @returns {Promise<Object>} On success `{ namespaces: Array<string> }`;
  *  on error `{error: string}`.
  */
-const _fetchNamespaces = async function (cloud, loadAll) {
+const _fetchNamespaces = async function (cloud) {
   // return syncNamespaces if we already define them
-  if (!loadAll && cloud?.syncNamespaces?.length) {
+  if (cloud?.syncNamespaces?.length) {
     return { namespaces: cloud.syncNamespaces };
   }
   const { error, body } = await authedRequest({
@@ -440,18 +439,23 @@ export class ExtendedCloud {
     }
   }
 
-  /**
-   *
-   * @param {boolean} [loadAll] fetch all namespaces, ignore syncNamespaces
-   * @return {Promise<ExtendedCloud>}
-   */
-  async init(loadAll) {
+  async init() {
     this.loading = true;
     this.error = null;
+    // if no config (eg when we restore cloud from disk) try to load it
+    if (!this.cloud?.config) {
+      await this.cloud.loadConfig();
+      // if loadConfig error we get it as connectError
+      // in this case stop init and return ExtendedCloud with that error
+      if (this.cloud.connectError) {
+        this.error = getErrorMessage(this.cloud.connectError);
+        this.loading = false;
+        return this;
+      }
+    }
 
     const { error: nameSpaceError, namespaces } = await _fetchNamespaces(
-      this.cloud,
-      loadAll
+      this.cloud
     );
     if (nameSpaceError) {
       this.error = getErrorMessage(nameSpaceError);
@@ -503,8 +507,8 @@ export class ExtendedCloud {
         });
 
         this.loading = false;
-        return this;
       }
     }
+    return this;
   }
 }
--- a/src/renderer/store/ExtendedCloudProvider.js
+++ b/src/renderer/store/ExtendedCloudProvider.js
@@ -60,8 +60,8 @@ const _loadData = function (tokens, cloudUrlsToUpdate) {
     const extCl = new ExtendedCloud(cloud);
     // if token isn't valid we don't need to add listeners, etc
     // because on next step it has to be reconnected and rewritten anyway
-    // but we gave to add it no store.extendedClouds all cases, to show in the table
-    if (!extCl.cloud.token || extCl.cloud.isRefreshTokenExpired()) {
+    // but we have to add it to store.extendedClouds all cases, to show in the table
+    if (!extCl.cloud.isValid()) {
       pr.store.extendedClouds[url] = extCl;
     } else {
       extCl.addEventListener(LOADING_CHANGE, updateSingleCloud);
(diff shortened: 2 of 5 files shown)


src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,140 @@
...
+    );
+    extCl.startUpdateCloudByTimeOut();
+    pr.store.extendedClouds[url] = extCl;
+    // we need to know when all async promises are finished
+    // so increase count and when it equal cloudUrls.length set valid loading/loader statuses
+    loadFinishedCount += 1;
+    if (cloudUrls.length === loadFinishedCount) {
+      pr.loading = false;
+      pr.loaded = true;
+    }
```
stefcameron, line 87: I don't think this is working as you're thinking. `pr.onChange()` will be called as soon as your loop is done here, but extended cloud objects will still be loading. `new ExtendedCloud(cloud)` will trigger the initial load in the next frame.

The only way to know when it's done is to add a listener for the LOADING_CHANGED event. So I would expect to see listeners added, and the `loadFinishedCount` counter increased for every call to the listener until the total is reached. Then you can set `pr.loading = false; pr.loaded = true; pr.onChange();`.

And I think we would never set `pr.error` because each ExtendedCloud has its own `error` property that users of `useExtendedCloud()` can check, so might be good to indicate that expectation with a comment so we remember we thought about `pr.error` and didn't just forget to set it to something.

stefcameron (commented):
This is looking good! 😄

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,140 @@
...
+  const [state, setState] = useState(pr.clone());
+  const value = useMemo(() => [state, setState], [state]);
+
+  pr.setState = setState;
+  // autorun calls on any cloudStore update (to often)
+  autorun(() => {
+    const cloudsNeedToBeUpdated = [];
+    // we need to take stable part to compare and observe
+    // cloud.token, as I understand, is good for that
+    // when we update cloud - we update its token (for now at least, this 'compare' part may be changed in the future)
```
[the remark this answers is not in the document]
bnalisnikovskiy: In this case `token` should suite well. We need to update (recreate) UI extendedCloud when cloud on disk is updated. Url stays the same, but we can't store cloud without `token` changes

src/renderer/store/ExtendedCloudProvider.js
bnalisnikovskiy: Yes, you are right. I had added this logic when we had `await extCloud.init()` separately, and in that case it worked well. I'll improve this

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,146 @@
...
+  pr.loaded = false;
+  pr.onChange();
+
+  pr.store.tokens = tokens;
+  cloudUrls.forEach(async (url) => {
+    const cloud = cloudStore.clouds[url];
+    await cloud.loadConfig(); // we need to get config for getting namespaces and other at ExtendedCloud.init()
+    const extCl = new ExtendedCloud(cloud);
+
+    extCl.addEventListener(DATA_UPDATED, updateSingleCloud);
```
stefcameron, line 81: Because we just created the ExtendedCloud instance, we should be listening for the LOADING_CHANGED event, not DATA_UPDATED, so that we don't have added complexity of having to handle an as-yet unloaded EC object in `updateSingleCloud()`. Right now, it feels like there's still a hole in `updateSingleCloud()` when the EC isn't loaded yet.

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,146 @@
...
+  pr.store.loadFinishedCount = 0;
+  pr.store.cloudsToUpdateCount = cloudUrls.length;
+  pr.loading = true;
+  pr.loaded = false;
+  pr.onChange();
+
+  pr.store.tokens = tokens;
+  cloudUrls.forEach(async (url) => {
+    const cloud = cloudStore.clouds[url];
+    await cloud.loadConfig(); // we need to get config for getting namespaces and other at ExtendedCloud.init()
```
stefcameron, line 78: What if this promise fails? It'll cause the outer promise to fail, and that will cause an exception that will take down the whole app because it'll be an "unhandled rejected promise". I think it would also cause issues with the outer loop, possibly causing some or all other ExtendedClouds to fail to be created.

Also, calling `cloud.loadConfig()` here also feels like a detail that doesn't belong here where we're dealing with ExtendedClouds not Clouds. I think `ExtendedCloud#init()` should, internally, call `await this.cloud.loadConfig()` (with a `try/catch` in case it fails), which, in turn, would internally check if it's config is already loaded and just do nothing, or actually load the config.

Finally, I noticed `ExtendedCloud#init()` has a `loadAll` parameter: Do we need that parameter? Given we create an `ExtendedCloud` with a `Cloud` object, that that the `Cloud` instance has a property that tells us if all namespaces are synced, or only select ones, shouldn't what we load in `EC#init()` be dependent upon `this.cloud.syncAll` instead? Besides, when we create a new EC instance, we call `setTimeout(() => this.init());` which means we always initialize with `loadAll === undefined`, which isn't correct.

stefcameron (changes requested):
This is going in the right direction, but I don't think those temporary counters should be getting persisted to disk...

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,146 @@
...
+
+class ExtendedCloudProviderStore extends ProviderStore {
+  // @override
+  makeNew() {
+    return {
+      ...super.makeNew(),
+      extendedClouds: {},
+      tokens: null,
+      loadFinishedCount: 0,
+      cloudsToUpdateCount: 0,
```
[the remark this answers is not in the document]
bnalisnikovskiy: We are not storing them on the disk. `ExtendedCloudProviderStore` is local (UI) storage. Probably we won't need this at all, not sure if general `loading`, `loaded` will be useful, since we observe each ExtendedCloud and it has its own loader

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,146 @@
...
+ * @param {Array<string>} cloudUrlsToUpdate, list of cloudUrls that has to be updated
+ * @private
+ */
+const _loadData = function (tokens, cloudUrlsToUpdate) {
+  const cloudUrls = cloudUrlsToUpdate?.length
+    ? cloudUrlsToUpdate
+    : Object.keys(cloudStore.clouds);
+
+  pr.store.loadFinishedCount = 0;
+  pr.store.cloudsToUpdateCount = cloudUrls.length;
```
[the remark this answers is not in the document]
bnalisnikovskiy: We don't persist them. It's only UI part to inform us when all `cloudUrlsToUpdate` are loaded

src/renderer/store/ExtendedCloudProvider.js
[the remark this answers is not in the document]
bnalisnikovskiy: Update: I decided to remove them at all. I don't think we need 'loaders' for full clouds list. If we want we can listen each cloud separately and show loaders for particular ones

src/renderer/store/ExtendedCloudProvider.js
bnalisnikovskiy: It depends what we want. As we discussed [here](https://github.com/Mirantis/lens-extension-cc/pull/671#discussion_r788930269) in this place I follow `namespaces` updates only (without loading changes). 
However, if we want to show loader, better to use `LOADING_CHANGED`,  you are right

src/renderer/store/ExtendedCloudProvider.js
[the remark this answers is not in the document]
stefcameron: You're right, my mistake in thinking these were getting stored to disk. Sorry for that confusion. But it sounds like you've removed them and the loading/loaded/error too, which I think is good.

src/renderer/store/ExtendedCloudProvider.js
stefcameron: In my mind, because we're adding the event listener right after creating the EC instance, and then as soon as we're notified, we remove it, what we're really listening for is the initial load, which is `LOADING_CHANGED`. That was more intuitive to me than `DATA_UPDATED`, though I suppose the end-result, from the provider's perspective, is probably the same with either event.

bnalisnikovskiy: I tried both and `LOADING_CHANGED` works better.

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,133 @@
...
+
+const pr = new ExtendedCloudProviderStore();
+
+const ExtendedCloudContext = createContext();
+
+const updateSingleCloud = (extendedCloud) => {
+  const { cloudUrl } = extendedCloud?.cloud || {};
+  if (cloudUrl) {
+    pr.store.extendedClouds[cloudUrl] = extendedCloud;
+    pr.onChange();
```
stefcameron, line 42: Do you think that at this point, we should now remove the LOADING_CHANGED listener, and add a listener for DATA_UPDATED? That listener would be `updateSingleCloud()` again. Just now triggering a context update (via `pr.onChange()`) to the ExtendedCloudProvider whenever an EC is updated. Or maybe this will result in too context updates (and so reloads of the SyncView), which would hurt performance, and so the SyncView should be the one to subscribe to individual ECs?

stefcameron (approved):
Looking good! Some questions, but I don't think they're blocking issues. They could be handled in a subsequent PR, or here, or maybe you have different opinion. 🙂

src/renderer/store/ExtendedCloudProvider.js
bnalisnikovskiy: If we want to show loader for single clouds, we should use `LOADING_CHANGED`. Technically, we don't rewrite Cloud (ref leads to the same object, but I'l double check) - so we shouldn't remove/update listeners. Or, maybe, I don't fully understand the comment...

stefcameron: The side effect of `pr.onChange()` is that it will cause a change to the value provided by the context provider. That means that any piece of the UI that calls `useExtendedClouds()` to get the list of EC objects will get re-rendered.

The way you've coded this right now, because ECs load async over time, let's say we have 4 ECs, there will likely be 4 re-render cycles of all components that `useExtendedClouds()` as each EC finishes loading (in the case where they just end-up finishing serially, one after another).

That might negatively affect performance. If the list of ECs has not changed, we shouldn't `pr.onChange()`. We should only `pr.onChange()` when the list changes (an EC is added or removed completely).

So `pr.onChange()` should be called in `_loadData()` __once__ after the `cloudUrls` have been processed to notify users of the context provider that the list has changed in some significant way. That means that this statement should move down, just __after__ the `cloudUrls.forEach()` loop: https://github.com/Mirantis/lens-extension-cc/pull/683/files#diff-5ab31a3171545a614baaf23b1e4a69562bdc11a96a322c54e42c1ff591951f02R81

Then each view that `useExtendedClouds()` will get updated only when there's a significant change to the list of ECs that it provides, and from there, each view is responsible to add its own event listeners (for LOADING_CHANGE and/or DATA_UPDATED) on each EC in order to show a loader and update its UI.

In this way, we have a single provider for all ECs, but then each EC is treated individually.

bnalisnikovskiy: Okay. Let's think how it would work. We can call pr.onChange() in `_loadData` below forEach loop. 
`_loadData` happens in two cases: Initial load, and when `token` in Clouds is updated (our `updateTokens()` call). But if in ExtedndedCloud.init() method (each 5 min) we get new namespaces, with different count of sshKeys/credentials/etc `_loadData` won't be called. 
We can add listeners on component level (if I got your idea correctly) for each ExtendedCloud and show loader etc, but in this case we will have two source of truth. ExtendedClouds from `ExtendedCloudProvider` (outdated) and particular ExtendedCloud in component with new data. So here two ways: we need to update `ExtendedCloudProvider` from component somehow (and use it as one source of truth) or wait until new data come with next `_loadData` call (it happens each ~20 min, for one cloud, as I can see). 
 I don't think that's good approach

About possible performance issues. React optimized its strategy and comparison mechanism very good. If keys stay the same rendering takes very small time, we shouldn't afraid of that. Also we could look to some libraries, like 'reselect', to have memorized data, etc

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,133 @@
...
+    : Object.keys(cloudStore.clouds);
+
+  pr.store.tokens = tokens;
+  cloudUrls.forEach(async (url) => {
+    const cloud = cloudStore.clouds[url];
+    const extCl = new ExtendedCloud(cloud);
+    // if token isn't valid we don't need to add listeners, etc
+    // because on next step it has to be reconnected and rewritten anyway
+    // but we gave to add it no store.extendedClouds all cases, to show in the table
+    if (!extCl.cloud.token || extCl.cloud.isRefreshTokenExpired()) {
```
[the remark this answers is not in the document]
bnalisnikovskiy: In `isValid` we also check `username`. If it's a required field and it always has to be setted, than `isValid` suits better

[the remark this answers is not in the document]
stefcameron: We might be able to remove `username` from the check. It's not required in terms of being "connected" because all the API calls are done using tokens, not username/password with basic auth.

The `username` property is required (i.e. very important), however, for purposes of creating kubeconfig files that Lens will consume in order to add clusters.

I think `isConnected()` would be more intuitive when reading the code, and we can remove the `username` check inside of it so that the check is purely related to having tokens that are valid.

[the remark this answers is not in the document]
stefcameron: I don't think we need a separate method. I think we can just rename `isValid` to `isConnected` and remove the `username` check inside of it. I was just explaining how the `username` is still an important piece of data, but it's not necessary in terms of connecting to the Cloud via API, so doesn't need to be in that check anymore. It's a relic from when this was the `AuthAccess` class.

src/renderer/store/ExtendedCloudProvider.js
stefcameron: Can you point me to docs on `reselect`? I don't know about it (or don't remember it).

bnalisnikovskiy: https://github.com/reduxjs/reselect

stefcameron: > I think I might create selector which takes 'url' and returns only one cloud when it's changed. So we might listen separately each cloud in their own 'row' components instead of connecting it through `useExtendedCloudData` on parent.

Will that also address the scenario where the number of ECs changes?.

The SyncView would do something like this with the provider:

```javascript
const { extendedClouds } = useExtendedClouds();
...
return (
  <List>
    {Object.values(extendedClouds).map((ec) => <SynchronizeBlock extendedCloud={ec} />)}
  </List>
);
```

Now you could use `reselect` and create a selector that reacts to changes to individual items within that map, so that would result in each SynchronizeBlock updating itself whenever the token expires and we get a new one. At the same time, though, `_loadData()` will run and will call `pr.onChange()` and then the SyncView itself will re-render, re-creating all the SynchronizeBlock instances anyway. So what benefit is there to the selector?

I don't see how we can improve performance other than to create an optimization in `_loadData()` where we don't create new EC instances if we already have them, but rather get them to update their data.

In the case where there are no new Clouds, and no old Clouds removed (the code isn't covering this case, BTW), but just new tokens for existing Clouds, ECs should just update their data, and SynchronizeBlock component instances should be listening to their respective EC's DATA_UPDATED event and refresh the UI when data changes. `_loadData()` should not call `pr.onChange()` in this case, and this will avoid SyncView getting re-rendered without cause.

There's still just one source of truth there about what the ECs are: The ExtendedCloudProvider.

In the case where a new Cloud is added, or an existing one is removed -- which means the number of ECs will change -- then `_loadData()` needs to detect that and call `pr.onChange()`. Also, in the case where a Cloud was removed, we don't need to call `_loadData()` (nothing to load). We just need to `delete pr.store.extendedClouds[url]` and remove the token also, and then call `pr.onChange()`.

Adding/removing Clouds should be an infrequent action. Clouds and ECs getting updated will happen all the time. We should make sure that's optimized.

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -0,0 +1,133 @@
...
+ * @param {Array<string>} cloudUrlsToUpdate, list of cloudUrls that has to be updated
+ * @private
+ */
+const _loadData = function (tokens, cloudUrlsToUpdate) {
+  const cloudUrls = cloudUrlsToUpdate?.length
+    ? cloudUrlsToUpdate
+    : Object.keys(cloudStore.clouds);
+
+  pr.store.tokens = tokens;
+  cloudUrls.forEach(async (url) => {
```
stefcameron, line 58: Where are we handling the case where a Cloud has been removed? AFAICT, a removed Cloud would get removed from the `cloudStore` but would remain forever in the `extendedClouds` map as an EC instance that would keep getting rendered. We're missing a `delete pr.store.extendedClouds[url]` somewhere...

src/renderer/store/ExtendedCloudProvider.js
bnalisnikovskiy: I'll make a test branch tomorrow and show what I mean. In my view, with that optimization we almost shouldn't change a logic. Only in remove part, you are right about it, I need to test that better.

src/renderer/store/ExtendedCloudProvider.js
bnalisnikovskiy: When cloud is removed it triggers `_loadData` and we load only existing clouds to ECProvider. But you're right in point that we don't need load data for clouds which left. I'll figure out how to do that in efficient way and avoid extra calls. Good point

stefcameron (approved):
The latest updates look good. We just need to optimize the case where an existing Cloud (for which we already have an EC in the provider) is updated: Update the existing EC (eventually resulting in the EC firing its DATA_UPDATED event) rather than create a new instance, and don't fire `pr.onChange()` if all that was done in `_loadData()` was to update existing ECs. Also don't fire `pr.onChange()` whenever an EC has been updated. Let each "consumer" of an EC add its own DATA_UPDATED and/or LOADING_CHANGED listener, as appropriate.

You can optimize this in a follow-up PR if you want. This PR should work as-is for now.

src/common/ExtendedCloud.js
```diff
@@ -384,11 +385,16 @@ export class ExtendedCloud {
...
 
     Object.defineProperty(this, 'cloud', {
       enumerable: true,
       get() {
-        return cloud;
+        return _cloud;
+      },
+      set(newValue) {
+          _cloud = newValue;
+          this.dispatchEvent(EXTENDED_CLOUD_EVENTS.DATA_UPDATED, this);
```
stefcameron: I don't think we should be firing this event when the Cloud itself changes. The data hasn't changed at this point. We should be calling `this.init()` -- which we should rename to something like `updateData()` or `fetchData()` because it's weird to be calling `init()` every 5 minutes to do a data update... We're no longer initializing anything at that point.

stefcameron (approved):
Looks good again!

PR #690, opened 2022-01-25
PRODX-21255 Update clouds on disk after any tokens update

bnalisnikovskiy:


src/common/Cloud.js
```diff
@@ -656,6 +669,14 @@ export class Cloud {
     };
   }
 
+  async loadConfig() {
+    try {
+      this.config = await _loadConfig(this.cloudUrl);
```
stefcameron, line 674: Do you think it would be a good optimization to first check if `this.config` is defined, and if so, to just return instead of actually loading the config again? Or are we better off always loading the config? Seems like with what's going on in ExtendedCloud in #683, it might be good for performance to have this optimization.

stefcameron (approved):
This looks good to me! Approved, but please consider that one question before merging.

src/common/Cloud.js
bnalisnikovskiy: Good point. I'll optimise this in [#683](https://github.com/Mirantis/lens-extension-cc/pull/683)

PR #692, opened 2022-01-25
[PRODX-21176] Enable opening clusters in Lens based on context

stefcameron:
When manually adding a cluster from MCC into Lens via downloaded
kubeconfig, Lens generates its own UID for the cluster since the
UID isn't given in the kubeconfig. That means we can't search
by UID when opening a cluster from MCC in the browser that
wasn't added via the extension.

This changes to searching by the same context name that MCC generates
in its kubeconfigs. So now we can find any cluster, whether it was
manually added, or added via the extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # ChangeLog
 
-## UNRELEASED
+## v3.0.4
 
 ### Patch
 
--- a/src/common/Cloud.js
+++ b/src/common/Cloud.js
@@ -4,11 +4,11 @@ import { logger } from '../util/logger';
 import * as strings from '../strings';
 import * as ssoUtil from '../util/ssoUtil';
 import {
+  EXT_EVENT_OAUTH_CODE,
   extEventOauthCodeTs,
   addExtEventHandler,
   removeExtEventHandler,
 } from '../renderer/eventBus';
-import { EXT_EVENT_OAUTH_CODE } from '../constants';
 
 /**
  * Determines if a date has passed.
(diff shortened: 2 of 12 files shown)


stefcameron:
Note I also fixed-up a couple of mix-ups that happened related to the "activate cluster" protocol event during the recent clean-up (i.e. removal of "add to lens" support) and consolidated all things related to extension events into the eventBus.

stefcameron:
Also note this "ports" the feature originally added to the v3 maintenance branch into the new code: https://github.com/Mirantis/lens-extension-cc/pull/691

bnalisnikovskiy (changes requested):
It looks good. Only one question about `EXT_EVENT_OAUTH_CODE`
We moved it from EventBus to avoid  circular dependency: eventBus -> Cloud -> ssoUtil -> AuthClient -> eventBus 
Check please this moment
image

stefcameron:
> It looks good. Only one question about `EXT_EVENT_OAUTH_CODE` We moved it from EventBus to avoid circular dependency: eventBus -> Cloud -> ssoUtil -> AuthClient -> eventBus Check please this moment We discussed it [here in Slack](https://miracloud.slack.com/archives/D02PHAV9FAQ/p1641487755036100)

Oh, yes, we did. But it seems the circular dependency isn't there anymore because I tested this in Lens. The extension loads fine. `eventBus` no longer has any imports except for `rtvjs`, so whatever it was importing from `Cloud`, it no longer needs.

bnalisnikovskiy (approved):

PR #693, opened 2022-01-26
[PRODX-20981]: added clusters table

AntonNos:


src/renderer/components/EnhancedTable/EnhancedTable.js
```diff
@@ -0,0 +1,110 @@
...
+
+function getComparator(order, orderBy) {
+  return order === 'desc'
+    ? (a, b) => descendingComparator(a, b, orderBy)
+    : (a, b) => -descendingComparator(a, b, orderBy);
+}
+
+// This method is created for cross-browser compatibility, if you don't
+// need to support IE11, you can use Array.prototype.sort() directly
+function stableSort(array, comparator) {
```
bnalisnikovskiy, line 32: we don't need IE support ;)

bnalisnikovskiy (changes requested):

src/renderer/components/EnhancedTable/EnhancedTable.js
stefcameron: IE 😱 !!!!

src/renderer/components/GlobalPage/ManagementClusters.js
```diff
@@ -0,0 +1,66 @@
+import styled from '@emotion/styled';
```
stefcameron, line 1: This component is what the existing `SyncView` should become. We should replace what's in `SyncView` with this instead of creating a new component here. Just delete everything that's in SyncView and replace it with this content -- except maybe https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer/components/GlobalPage/SyncView.js#L346-L349 if we still need a hacky way to show that wizard for the time being. FYI @bnalisnikovskiy

stefcameron (commented):
Also, we're adding a new dependency on `@material-ui` but this PR doesn't include changes to the package.json nor to the lock file.

src/renderer/components/EnhancedTable/EnhancedTable.js
AntonNos: Removed it:) Will add sorting in another PR, because we will have release in Feb, and I can't block Bogdan with this table

src/renderer/components/GlobalPage/ManagementClusters.js
stefcameron: @AntonNos Do you plan on putting this code into the SyncView? We don't need this ManagementClusters view. This _is_ what the SyncView is supposed to display. 🙂

bnalisnikovskiy: I'll do that in my next PR. I already have a workpiece for that. In SyncView we should use Welcome, Connection, pages as well. So better do that all at once

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -0,0 +1,148 @@
...
+        <EnhTableCell>STATUS</EnhTableCell>
+        <EnhTableCell isRightAligned>
+          <EnhMoreButton>
+            <Component.Icon material="more_vert" style={moreInfoIconStyles} />
+          </EnhMoreButton>
+        </EnhTableCell>
+      </EnhTableRow>
+      {isOpenFirstLevel &&
+        row.namespaces.map((namespace, index) => (
+          <div style={{ display: 'contents' }} key={namespace.name}>
```
bnalisnikovskiy, line 105: inline style

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -0,0 +1,148 @@
...
+import { AdditionalInfoRows } from './AdditionalInfoRows';
+
+const { Component } = Renderer;
+
+const EnhTableRow = styled.tr`
+  background-color: ${({ isTopLevel }) =>
+    isTopLevel ? 'var(--layoutTabsBackground)' : 'var(--mainBackground)'};
+`;
+
+const EnhTableCell = styled.td`
```
bnalisnikovskiy, line 15: I see that we have 4 `EnhTableCell` declaration in different files. Probably some of them could be renamed more specific...

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -0,0 +1,148 @@
...
+                    <Component.Icon
+                      material="chevron_right"
+                      style={expandIconStyles}
+                    />
+                  )}
+                </EnhCollapseBtn>
+                {namespace.name}
+              </EnhTableCell>
+              <EnhTableCell></EnhTableCell>
+              <EnhTableCell></EnhTableCell>
```
bnalisnikovskiy, line 124: just FYI, we can use shorter form `<EnhTableCell />` if component is empty. Eslint even has rule for that, obviously, we have it turned off 
No changes is required here. Up to you

bnalisnikovskiy (changes requested):
Looks good, but need a few improvements

bnalisnikovskiy (approved):
Looks good. I think It's ready to merge. Other additional features (like sorting) or possible bugs may be fixed in separate tickets

PR #695, opened 2022-01-28
[PRODX-21380]: added sorting to table

AntonNos:


stefcameron:
@AntonNos I looked at the code, but there seems to be a lot of code I've also seen in #693. Is this PR correctly updated? It seems like it should currently be based on (i.e. set to merge into) the #693 branch, not `master`?

stefcameron:
Like the history of this PR alone appears to include all changes to the repo since Dec ~17 or so?

stefcameron:
Let's review this over on #696 . It's identical to this PR except with a clean git history.

PR #696, opened 2022-01-28
[PRODX-21380] Adding sort feature to table

stefcameron:
This PR is identical to #695 except is has a clean git history.

### DIFF
--- a/src/renderer/components/EnhancedTable/EnhancedTable.js
+++ b/src/renderer/components/EnhancedTable/EnhancedTable.js
@@ -1,4 +1,6 @@
 import PropTypes from 'prop-types';
+import { useState } from 'react';
+import _ from 'lodash';
 import styled from '@emotion/styled';
 import { layout } from '../styles';
 import { EnhancedTableHead } from './EnhancedTableHead';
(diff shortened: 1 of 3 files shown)


src/renderer/components/EnhancedTable/EnhancedTable.js
```diff
@@ -1,4 +1,6 @@
 import PropTypes from 'prop-types';
+import { useState } from 'react';
+import _ from 'lodash';
```
stefcameron, line 3: For Webpack optimization, it would be better to specifically import the functions we need instead of all of Lodash. So `import { get, orderBy } from 'lodash';` instead of this.

src/renderer/components/EnhancedTable/EnhancedTable.js
```diff
@@ -23,12 +25,52 @@ const emptyRowStyles = {
   backgroundColor: 'var(--mainBackground)',
 };
 
+export const HEAD_CELL_VALUES = {
+  NAME: 'Name',
+  URL: 'URL',
+  USERNAME: 'Username',
+  STATUS: 'Status',
```
stefcameron, line 32: These labels should be coming from the `strings.ts` module.

stefcameron: Sorry, I misunderstood the code. These aren't actually rendered.

src/renderer/components/EnhancedTable/EnhancedTable.js
AntonNos: Fixed

stefcameron:
@AntonNos These changes look good to me now, thank you! I'll merge this.

PR #697, opened 2022-01-31
PRODX-21405 Connect components together

bnalisnikovskiy:
image


stefcameron:
👀

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -59,11 +61,42 @@ const moreInfoIconStyles = {
...
+        cloudStatus: connectionStatuses.cloud.connected(),
+        namespaceStatus: connectionStatuses.namespace.connected(),
+      }
+    : {
+        cloudStatus: connectionStatuses.cloud.disconnected(),
+        namespaceStatus: connectionStatuses.namespace.disconnected(),
+      };
+};
+
 export const EnhancedTableRow = ({ row }) => {
```
stefcameron, line 80: Based on properties and methods called on `row`, it's an ExtendedCloud instance. Could we rename `row` to `extendedCloud` so it's clear what `row` is? `row`, as a name, doesn't indicate what it might be. The prop type is just `object`, so not even that would be a good hint of what `row` is expected to be.

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -77,18 +99,34 @@ export const ConnectionBlock = ({
...
-          checkError(newCloud);
+          checkConnectionError(newCloud);
         }
       }
     };
     newCloud.addEventListener(CLOUD_EVENTS.STATUS_CHANGE, statusListener);
     await newCloud.connect();
   };
 
+  const seName = (value) => {
```
stefcameron, line 112: Typo: `seName` -> `setName`

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -52,21 +52,43 @@ export const ConnectionBlock = ({
...
     if (cloud?.connectError) {
       Notifications.error(cloud.connectError);
     }
   };
 
+  const validateCloud = (url) => {
+    if (cloudStore?.clouds?.[url]) {
+      const errorMessage = connectionBlock.notice.urlAlreadyUsed();
+      setValidationError(errorMessage);
+      Notifications.error(errorMessage);
```
stefcameron, line 68: For a validation error, I think showing the error inline is enough. We don't need a notification message on top of it.

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -52,21 +52,43 @@ export const ConnectionBlock = ({
...
+      return false;
+    }
+    if (!clusterName.trim()) {
+      const errorMessage = connectionBlock.notice.nameIsEmpty();
+      setValidationError(errorMessage);
+      Notifications.error(errorMessage);
```
stefcameron, line 74: For a validation error, I think showing the error inline is enough. We don't need a notification message on top of it.

src/renderer/components/GlobalPage/GlobalPage.tsx
```diff
@@ -76,15 +76,13 @@ export const GlobalPage = function ({ extension }: Props) {
...
-            <SsoAuthProvider>
-              <ClusterDataProvider>
-                <ClusterActionsProvider>
-                  <SyncView />
-                </ClusterActionsProvider>
-              </ClusterDataProvider>
-            </SsoAuthProvider>
-          </ConfigProvider>
+          <ClusterDataProvider>
+            <ClusterActionsProvider>
```
stefcameron, line 80: These two aren't needed anymore, are they?

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -170,28 +171,51 @@ export const SynchronizeBlock = ({ extCloud }) => {
...
+
+    if (allNamespaces.length === namespaces.length) {
+      cloud.syncNamespaces = [];
+      cloud.syncAll = true;
+    } else {
+      cloud.syncAll = false;
+      cloud.syncNamespaces = namespaces;
+    }
+
+    cloudStore.addCloud(cloud);
```
stefcameron, line 215: Shouldn't happen here. I see SynchronizeBlock as a component, and this as business logic, that doesn't belong here. It belongs in the SyncView instead.

src/strings.ts
```diff
@@ -117,6 +53,8 @@ export const connectionBlock: Dict = {
   notice: {
     info: () =>
       "You will be directed to your Management Cluster's login page through your web browser where you should enter your SSO credentials",
+    urlAlreadyUsed: () => 'Management Cluster URL is already used',
```
stefcameron, line 56:
```suggestion
    urlAlreadyUsed: () => 'This Management Cluster is already being synced',
```

src/strings.ts
```diff
@@ -128,6 +66,9 @@ export const synchronizeBlock = {
     sshKeys: () => 'SSH keys',
     credentials: () => 'credentials',
   },
+  error: {
+    noProjects: () => 'You have to select at least one project',
```
stefcameron, line 70:
```suggestion
    noProjects: () => 'Select at least one project to sync',
```

src/strings.ts
```diff
@@ -213,24 +109,6 @@ export const addClusters: Dict = {
...
-  },
-  offline: {
-    label: () => 'Offline use',
-    tip: () =>
-      'Generating tokens for offline use is less secure because they will never expire',
-  },
-  saved: () => 'Preferences saved!',
-};
-
 export const clusterDataProvider: Dict = {
```
stefcameron, line 112: This can be deleted. So should `ClusterDataProvider.js`

stefcameron (changes requested):
Great to see this wholesale change to the new design and flow, and connecting everything!

src/renderer/components/GlobalPage/SyncView.js
```diff
@@ -1,413 +1,96 @@
...
+  const openAddCloudBlock = () => setShowAddCloudComponent(true);
+  const closeAddCloudBlock = () => setShowAddCloudComponent(false);
+
+  // we control this state and should check it first
+  if (showAddCloudComponent) {
+    return <AddCloudInstance closeAddCloudBlock={closeAddCloudBlock} />;
+  }
+  // in no clouds => show Welcome page
+  if (!Object.keys(cloudStore.clouds).length) {
+    return <WelcomeView openAddCloudBlock={openAddCloudBlock} />;
```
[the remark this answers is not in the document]
bnalisnikovskiy: They have different styles and condition to show. 
I can use `  const [showAddCloudComponent, setShowAddCloudComponent] = useState(!Object.keys(cloudStore.clouds).length);` 
and then move `WelcomeView` to `AddCloudInstance` but I want to keep `WelcomeView` as a separate component anyway. Don't want to move those lines (styles and wrappers) and 'dilute' AddCloudInstance. We will look into AddCloudInstance much more often than in WelcomeView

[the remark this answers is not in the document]
stefcameron: OK, I see your point about combining the components. My request about the event props remains, and would now apply to both WelcomeView and AddCloudInstance components.

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -52,21 +52,47 @@ export const ConnectionBlock = ({
...
     }
   };
 
+  const validateCloud = (url) => {
+    if (cloudStore?.clouds?.[url]) {
+      setValidationError(connectionBlock.notice.urlAlreadyUsed());
+      return false;
+    }
+    const validName = clusterName.trim();
+    if (!validName || validName.includes(' ')) {
```
stefcameron, line 70: Could contain a tab character too, if it was copy/pasted from somewhere else that allowed typing a tab character. I think this would cover all possible cases:

```suggestion
    if (!validName || validName.match(/\s/)) {
```

src/renderer/components/GlobalPage/GlobalPage.tsx
```diff
@@ -76,15 +74,9 @@ export const GlobalPage = function ({ extension }: Props) {
     <CacheProvider value={createCache({ key: 'lens-extension-cc' })}>
       <ThemeProvider theme={theme}>
         <ExtStateProvider extension={extension}>
```
stefcameron, line 76: I forgot about this one too, and related ClusterStore.js and PreferenceStore.js. I don't think we need any of those anymore either.

src/strings.ts
```diff
@@ -117,6 +53,10 @@ export const connectionBlock: Dict = {
   notice: {
     info: () =>
       "You will be directed to your Management Cluster's login page through your web browser where you should enter your SSO credentials",
+    urlAlreadyUsed: () => 'This Management Cluster is already being synced',
+    nameIsEmpty: () =>
+      'Management Cluster name cannot be empty or contain whitespaces',
```
stefcameron, line 58:
```suggestion
      'Management Cluster name cannot be empty or contain whitespace',
```

("whitespace" without the "s" works for both singular and plural cases)

src/strings.ts
```diff
@@ -117,6 +53,10 @@ export const connectionBlock: Dict = {
...
+    nameAlreadyUsed: () => 'This Management Cluster name is already exists',
```
stefcameron, line 59:
```suggestion
    nameAlreadyUsed: () => 'A Management Cluster with this name is already being synced',
```

src/strings.ts
```diff
@@ -172,34 +98,6 @@ export const configProvider: Dict = {
...
-    selectAll: {
-      label: () => 'Select all',
-    },
-    selectNone: {
-      label: () => 'Select none',
-    },
-  },
-};
-
 export const addClusters: Dict = {
```
stefcameron, line 101: This can be deleted

stefcameron (approved):
Thanks for the updates! There are a couple of small things remaining, but no need to review further. Just please update before you merge to master.

src/renderer/components/GlobalPage/GlobalPage.tsx
bnalisnikovskiy: We use `ExtStateProvider` in `ClusterPage`. And `ClusterPage` in `renderer.tsx` 
image
Should I remove all related to `ClusterPage` too?

src/strings.ts
bnalisnikovskiy: Thank you. I don't know many thing native-speakers know. So, very appreciated.

src/renderer/components/GlobalPage/GlobalPage.tsx
stefcameron: Good question. No, let's keep that. ClusterView is for a custom panel in the Lens Cluster view (not sure what else to call it; the UI you see when you open a cluster in Lens). It's wrapped in ExtStateProvider, but doesn't use it, so keep ClusterPage and everywhere it's hooked up, just remove ExtStateProvider from around it in `ClusterPage.tsx`

src/strings.ts
stefcameron: My pleasure!

PR #701, opened 2022-02-01
[PRODX-21409]: added selective sync table

AntonNos:


stefcameron:
@AntonNos @bnalisnikovskiy When you guys make a UI change, could you please try to remember to include a screenshot or even short recording of the changes/new behavior in the PR?

src/renderer/components/GlobalPage/SyncView.js
```diff
@@ -75,23 +84,45 @@ export const SyncView = () => {
...
-          <Button
-            variant="outlined"
-            label={managementClusters.syncButtonLabel()}
-          />
+          <TopButtonsWrapper>
+            {isSelectiveSyncView ? (
+              <>
+                <CancelButton
+                  plain
+                  label={managementClusters.cancelButtonLabel()}
```
stefcameron, line 92: These strings are related to the `SyncView`, not `ManagementClusters`... Same for `managementClusters.title()` above (line 86) and a few other strings below.

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -290,10 +219,16 @@ export const SynchronizeBlock = ({ extCloud, onAdd }) => {
 };
 
 SynchronizeBlock.propTypes = {
-  extCloud: PropTypes.object,
+  extendedCloud: PropTypes.object,
   onAdd: PropTypes.func.isRequired,
+  checkBoxChangeHandler: PropTypes.func,
+  parentCheckboxValue: PropTypes.string,
+  childrenCheckboxValue: PropTypes.func,
```
stefcameron, line 226: Same issue here with these props. They seem too specific to internal implementation. "Separation of concerns" may lead to what feels like a lot of repetition or "prop drilling", but a context provider can help with that (if necessary) and the pattern just ensures that components truly are components and can be moved around with greater ease without breaking.

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -89,26 +89,19 @@ const SortButton = styled.button`
...
+export const SynchronizeBlock = ({
+  extendedCloud,
+  onAdd,
+  checkBoxChangeHandler,
+  parentCheckboxValue,
+  childrenCheckboxValue,
+}) => {
   // @type {object} sorted object of projects
-  const [projectsList, setProjectsList] = useState(extCloud.namespaces);
+  const [projectsList, setProjectsList] = useState(extendedCloud.namespaces);
```
stefcameron, line 100: We should load the state with a shallow clone of `namespaces` so we don't risk affecting the original array that the EC has. `[...extendedCloud.namespaces]`

Also, `extendedCloud.namespaces` might be `null` here, but I think the Spread Operator will handle that gracefully and produce an empty array.

src/renderer/components/HOC/hocWithCheckboxes.js
```diff
@@ -0,0 +1,95 @@
...
+    ...Component.props,
+    checkBoxChangeHandler,
+    parentCheckboxValue: getParentCheckboxValue(),
+    childrenCheckboxValue,
+    withCheckboxes: true,
+  });
+};
+
+HOCWithCheckboxes.propTypes = {
+  Component: PropTypes.node.isRequired,
```
stefcameron, line 94: I think the proper type is `element`. `node` could be just a string or a number. Also, there's an expectation of props that this component will have. Please mention them here in a JSDoc.

src/renderer/components/HOC/hocWithCheckboxes.js
```diff
@@ -0,0 +1,95 @@
...
+import PropTypes from 'prop-types';
+
+const checkboxesStateObj = (extCloud) => {
+  return (extCloud?.namespaces || []).reduce((acc, namespace) => {
+    acc[namespace.name] = false;
+    return acc;
+  }, {});
+};
+
+export const HOCWithCheckboxes = ({ Component }) => {
```
stefcameron, line 12: Let's have HOCs start with lowercase "hoc" since they aren't really components per se. They shouldn't take props either, if we're properly following [the pattern](https://reactjs.org/docs/higher-order-components.html).

It should be a function that takes a component to wrap and returns it "enhanced" with some common prop/state logic like you have in here.

The code that uses it should look like: `const ComponentWithCheckboxes = hocWithCheckboxes(Component);`

And then somewhere you render `<ComponentWithCheckboxes ... />`

src/strings.ts
```diff
@@ -76,6 +76,8 @@ export const synchronizeBlock = {
 export const managementClusters = {
   title: () => 'Mirantis Container Cloud management clusters',
   syncButtonLabel: () => 'Selective sync...',
+  cancelButtonLabel: () => 'Cancel',
+  synchroniseProjectsButtonLabel: () => 'Synchronise selected projects',
```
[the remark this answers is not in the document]
bnalisnikovskiy: Seems like a typo in figma 
image

src/renderer/components/GlobalPage/SynchronizeBlock.js
bnalisnikovskiy: Fixed. But spread cannot handle `null` or `undefined`. Will be crash if don't do like `[...(extendedCloud.namespaces || [])]`

src/renderer/components/HOC/hocWithCheckboxes.js
bnalisnikovskiy: I use custom react hood instead.

src/strings.ts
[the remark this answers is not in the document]
stefcameron: That's a typo in Figma, yes. Should be a "z".

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: Interesting. You're correct! That's a tricky thing with JavaScript as the spread operator in an object and function parameters handles `null` and `undefined` without throwing, but spread in an array crashes. Huh. Hopefully there's a good reason... 🙄

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -89,26 +90,19 @@ const SortButton = styled.button`
...
+export const SynchronizeBlock = ({ extendedCloud, onAdd }) => {
+  const { setCheckboxValue, getCheckboxValue } = useCheckboxes(
+    makeCheckboxesInitialState(extendedCloud)
+  );
 
-export const SynchronizeBlock = ({ extCloud, onAdd }) => {
   // @type {object} sorted object of projects
-  const [projectsList, setProjectsList] = useState(extCloud.namespaces);
+  const [projectsList, setProjectsList] = useState([
+    ...(extendedCloud.namespaces || []),
```
stefcameron, line 100: I think not knowing if `ExtendedCloud.namespaces` is null or an array is going to lead to this sort of code in many places, which is awkward. I'm going to make a PR to ExtendedCloud to change a bit of loading/fetching logic so that it's always safe to assume `ExtendedCloud.namespaces` is an array. It'll be empty if it hasn't loaded yet.

👉 Also, should this be `ExtendedCloud.syncedNamespaces`? It probably depends on whether we're in "Selective sync" mode or just the regular SyncView... In "Selective Sync" mode, we need to show all namespaces and check only the ones we're syncing (which could be all). Otherwise, we're in normal SyncView and we should only show the `syncedNamespaces`.

src/renderer/components/hooks/useCheckboxes.js
```diff
@@ -0,0 +1,106 @@
...
+      childrenCheckboxes.some((el) => el === false) &&
+      childrenCheckboxes.some((el) => el === true)
+    ) {
+      return checkValues.MIXED;
+    }
+    if (checkboxesState.parent) {
+      return checkValues.CHECKED;
+    } else {
+      return checkValues.UNCHECKED;
+    }
```
stefcameron, line 63: I think this might be a lint error?

```suggestion
    if (checkboxesState.parent) {
      return checkValues.CHECKED;
    }

    return checkValues.UNCHECKED;
```

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: See #722

stefcameron (changes requested):
Nice work on that, much appreciated! This is looking much tighter. 😄 Just a couple more things, but just small in comparison. We should be able to get this wrapped up tomorrow.

Also, please see #722 which is slightly related to this, meant to simplify some code downstream from ExtendedCloud objects.

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -180,10 +195,55 @@ export const EnhancedTableRow = ({ extendedCloud }) => {
...
+        <EnhTableRowCell />
+        <EnhTableRowCell>{namespaceStatus}</EnhTableRowCell>
+        <EnhTableRowCell isRightAligned>
+          <EnhMore onClick={() => toggleMenu(!onOpen)}>
+            <MenuActions onOpen={onOpen}>
+              {namespaceMenuItems.map((item) => {
+                return (
+                  <MenuItem
+                    key={`${namespace.name}-${item.name}`}
+                    onClick={item.onClick}
```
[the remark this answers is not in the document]
bnalisnikovskiy: Name is fixed
Lets add `Namespace` (and probably EC for some cases) in particular tickets

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -144,18 +155,24 @@ const namespaceMenuItems = [
...
+  const [actualNamespaces, setActualNamespaces] = useState([
+    ...(extendedCloud[usedNamespaces] || []),
+  ]);
   const [openedSecondLevelListIndex, setOpenedSecondLevelListIndex] = useState(
     []
   );
   const updateNamespaces = (updatedRow) => {
     if (updatedRow) {
-      setActualNamespaces(updatedRow.syncedNamespaces);
+      setActualNamespaces(updatedRow[usedNamespaces]);
```
bnalisnikovskiy, line 175: Probably here we should update all `namespaces`, and doesn't matter SyncView table or Synchronize table it is (`withCheckboxes: true/false`). What fo you think @stefcameron  ?

src/renderer/components/GlobalPage/SynchronizeBlock.js
bnalisnikovskiy: in this component we have to show all namespaces, but you are right about this logic. It's suitable for `EnhancedTableRow` , used it there

src/renderer/components/EnhancedTable/EnhancedTableRow.js
stefcameron: I'm missing a bit of context here, but what we want is the SyncView to show only the synced namespaces (`ExtendedCloud.syncedNamespaces`) and the SynchronizeView to show all namespaces (`ExtendedCloud.namespaces`) with only the synced ones checked (and then let the user update their selections).

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -144,18 +155,24 @@ const namespaceMenuItems = [
...
+  const usedNamespaces = withCheckboxes ? 'namespaces' : 'syncedNamespaces';
+
   const [onOpen, toggleMenu] = useState(false);
   const [isOpenFirstLevel, setIsOpenFirstLevel] = useState(false);
-  const [actualNamespaces, setActualNamespaces] = useState(
-    extendedCloud.syncedNamespaces
-  );
+  const [actualNamespaces, setActualNamespaces] = useState([
+    ...(extendedCloud[usedNamespaces] || []),
+  ]);
```
stefcameron, line 169:
```suggestion
  const [actualNamespaces, setActualNamespaces] = useState(extendedCloud[usedNamespaces]);
```

That's possible with #722...

src/renderer/components/EnhancedTable/EnhancedTableRow.js
stefcameron: Looking at the code changes now, I see what you did with `usedNamespaces`. I think it works as you have it.

src/renderer/components/EnhancedTable/EnhancedTableRow.js
bnalisnikovskiy: I'll update after you branch be merged

stefcameron (approved):
Looking good! Just one minor thing remains, though not critical.

src/renderer/components/EnhancedTable/EnhancedTableRow.js
stefcameron: It's merged!

PR #707, opened 2022-02-03
Fix CVE-2022-0484

stefcameron:
This addresses CVE-2022-0484 where `openExternal()` provided by
Lens is designed to open _anything_, which means even `file:` protocols
that would result in local code execution (if MCC URL is spoofed and has
a config.js that provides a `file:` URL for Keycloak, for example).

NOTE: This ports commit 23330ad9181022157ee51fedbdfb4d45b848cf49 on
`v3-maint` branch to `master`.

### DIFF
--- a/src/util/ssoUtil.js
+++ b/src/util/ssoUtil.js
@@ -1,9 +1,8 @@
-import { Common } from '@k8slens/extensions';
 import { logger } from './logger';
 import { extractJwtPayload } from '../renderer/auth/authUtil';
 import { AuthClient } from '../renderer/auth/clients/AuthClient';
-
-const { Util } = Common;
+import { openBrowser } from '../util/netUtil';
+import * as strings from '../strings';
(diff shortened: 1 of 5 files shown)


src/util/ssoUtil.js
```diff
@@ -13,15 +12,21 @@ const { Util } = Common;
  * @param {Object} options.config MCC Config object.
  * @param {string} [options.state] Any string that should be returned verbatim
  *  in a `state` request parameter in the redirect request.
+ * @throws {Error} If the MCC instance doesn't support SSO.
+ * @throws {Error} If the MCC instance has an illegal Keycloak URL.
```
bnalisnikovskiy, line 16: Didn't know that we can describe `throws` twice in JSDoc. Nice

bnalisnikovskiy (approved):
LGTM

PR #711, opened 2022-02-05
Optimize ExtendedCloud data fetches, fix bugs

stefcameron:
There was too much churn when fetching data for an ExtendedCloud:
Duplicated requests, multiple parallel data fetches for the same
EC, and too much back and forth between the CloudStore and the
ExtendedCloudProvider whenever a Cloud would get updated with
new tokens after they would get refreshed.

I fixed and changed a number of things. Here's a list of the most
prominent issues I addressed:

- addEventListener() in Cloud and EC was immediately calling the event
  handler instead of adding it to the list of handlers.
- Cloud didn't have enough events. Just having the one DATA_UPDATED
  event for all property changes was resulting in the EC using that
  Cloud to get notified (and trigger a data fetch) much too often.
  The EC only needs to fetch new data if the Cloud's sync-related
  properties change. Nothing else.
- Only the CloudStore is interested in all possible Cloud property
  changes so it can persist it to disk.
- When the EC's `cloud` property is updated from the
  ExtendedCloudProvider, the Cloud object actually isn't new, so
  there's no need to trigger a data fetch whenever the property
  is set. Now we check to see if the new value is actually
  different.
- When updating the EC's `cloud` property to a new Cloud, we
  weren't checking to see if we already had a Cloud object
  and removing listeners.
- Delaying the data fetch internally within the EC using the
  event dispatching mechanism is a good way to de-duplicate
  data fetch requests, rather than imperatively making them.
- Cloud.idpClientId was not needed, removed.
- The AddCloudInstance component was not properly destroying
  the temporary EC instance it was creating, which created a memory
  leak, and a random EC that kept fetching data after the
  component was unmounted ("destroyed" meaning event listeners
  on the Cloud removed, and interval stopped).
- In EC.fetchData(), we were continuing with requests to load
  clusters and ssh keys even though a previous request had
  failed, which would ultimately result in discarding the data
  and setting EC.error instead, so now we don't pursue additional
  unnecessary requests.
- authUtil module now returns an additional `tokensRefreshed`
  property in its payload, indicating whether tokens were
  refreshed as part of making the request. Turns out there's
  no optimization to make with this now that the EC doesn't
  care if the Cloud's tokens get updated, but this is handy
  nonetheless, so I left it in place.
- CloudStore was leaking Clouds when it would get reloaded
  from disk (Lens does sync to the disk every now and then,
  so even while Lens is running, we can get a random read
  from disk, causing us to overwrite the existing `clouds`
  map). We don't support this, so now we just skip resetting
  the `clouds` map when it's already set. If you want to
  mess with the JSON file on disk, quit Lens, mess with
  it, and then restart Lens.

### DIFF
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -38,7 +38,22 @@ export const EXTENDED_CLOUD_EVENTS = Object.freeze({
   FETCH_DATA: 'fetchData',
 });
 
-const FIVE_MIN = 300 * 1000;
+/**
+ * Milliseconds at which new data will be fetched from the Cloud.
+ *
+ * __AVOID__ exactly 5 minutes because API tokens expire after 5 minutes and need to be
+ *  refreshed. This will still work, but because we execute many parallel requests when
+ *  fetching data (e.g. for each namespace synced, multiple types of credentials), it'll
+ *  cause a lot of churn because each one of those will trigger a token refresh.
+ *
+ * Since fetching starts with getting all namespaces, and this is a single request, we
+ *  set this interval such that it's most likely this will be the request that causes
+ *  tokens to be refreshed, and then subsequent requests for remaining data (clusters,
+ *  SSH keys, etc.) will proceed with new, valid tokens.
+ *
+ * @type {number}
+ */
+const FETCH_INTERVAL = 4.85 * 60 * 1000; // 4:51 minutes (AVOID exactly 5 minutes)
 
 const credentialTypes = [
   'awscredential',
@@ -326,18 +341,8 @@ const _fetchClusters = async function (cloud, namespaces) {
   return dsError || { clusters, tokensRefreshed };
 };
 
-const getId = (function () {
-  let nextId = 0;
-  return () => {
-    nextId++;
-    return nextId;
-  };
-})(); // DEBUG REMOVE
-
 export class ExtendedCloud {
   constructor(cloud) {
-    this.INSTANCE_ID = getId(); // DEBUG REMOVE
-
     const _eventListeners = {}; // map of event name to array of functions that are handlers
     const _eventQueue = []; // list of `{ name: string, params: Array }` objects to dispatch
     let _dispatchTimerId; // ID of the scheduled dispatch timer if a dispatch is scheduled, or undefined
@@ -363,10 +368,6 @@ export class ExtendedCloud {
             events.forEach((event) => {
               const { name, params } = event;
               const handlers = _eventListeners[name] || [];
-              console.log(
-                '####### ExtendedCloud._scheduleDispatch(): DISPATCHING event=%s',
-                name
-              ); // DEBUG LOG
               handlers.forEach((handler) => {
                 try {
                   handler(...params);
@@ -382,10 +383,6 @@ export class ExtendedCloud {
             }
           }
         });
-      } else {
-        console.log(
-          '####### ExtendedCloud._scheduleDispatch(): already scheduled'
-        ); // DEBUG LOG
       }
     };
 
@@ -503,10 +500,6 @@ export class ExtendedCloud {
             `Received new Cloud: Scheduling new data fetch; extCloud=${this}`
           );
           this.dispatchEvent(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this);
-        } else {
-          console.log(
-            `####### ExtendedCloud.[set]cloud: NO CHANGE in cloud object`
-          ); // DEBUG LOG
         }
       },
     });
@@ -545,10 +538,6 @@ export class ExtendedCloud {
       set(newValue) {
         if (newValue !== _namespaces) {
           _namespaces = newValue || null;
-          console.log(
-            `####### ExtendedCloud.[set]namespaces: Set new namespaces data for extCloud=${this}, namespaces=`,
-            _namespaces
-          ); // DEBUG LOG
           this.dispatchEvent(EXTENDED_CLOUD_EVENTS.DATA_UPDATED, this);
         }
       },
@@ -589,7 +578,7 @@ export class ExtendedCloud {
       //  we don't duplicate an existing request to fetch the data if one has
       //  just been scheduled
       this.dispatchEvent(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this);
-    }, FIVE_MIN);
+    }, FETCH_INTERVAL);
 
     // schedule the initial data load/fetch
     this.dispatchEvent(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this);
@@ -597,8 +586,6 @@ export class ExtendedCloud {
 
   /** Called when this instance is being deleted/destroyed. */
   destroy() {
-    console.log(`####### ExtendedCloud.destroy(): DESTROYING extCloud=${this}`); // DEBUG LOG
-
     this.removeEventListener(
       EXTENDED_CLOUD_EVENTS.FETCH_DATA,
       this.onFetchData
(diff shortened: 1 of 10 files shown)


stefcameron:
__NOTE__: In case you guys ask me to make some changes, I left all my `console` debugging statements in place. 😄 

👉 All statements that have a `// DEBUG LOG` or `// DEBUG REMOVE` comment after them __will be removed__ before I merge. (They'll cause the build to break too, since `// DEBUG` results in a linting error.)

stefcameron:
And there is one issue that I've asked the Lens team for help on, which is that, at least for me, Lens is no longer writing data to the `cloud-store.json` file on disk, even though we're generating valid JSON for it. No idea why.

bnalisnikovskiy:
Seems we have a problem with store-on-disk. When I add new Cloud, after 'Synchronize Selected click' I redirected to SyncView and it appears in the table (all is good), but cloud-store.json doesn't have that cloud.

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -31,57 +31,100 @@ const pr = new ExtendedCloudProviderStore();
...
+    Object.keys(currentTokens).forEach((curUrl) => {
+      if (!storeTokens[curUrl]) {
+        // not a stored/known URL: must be new
+        newUrls.push(curUrl);
+      }
+    });
   }
-  return [];
+
+  return { newUrls, updatedUrls, oldUrls };
```
bnalisnikovskiy, line 70: I see a few gaps in this `_triageCloudUrls` function.  On line 50 you check  `if(currentTokens[storeUrl])` but it's  not a `updatedUrls` we push all clouds present in `storeTokens` to  `updatedUrls`. We need compare tokens inside and only then push url (if needed). I implemented that check on line 167 in `reduce` to avoid one more iteration.

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -31,57 +31,100 @@ const pr = new ExtendedCloudProviderStore();
...
+
+  const curCount = Object.keys(currentTokens).length;
+  if (updatedUrls.length !== curCount && oldUrls.length !== curCount) {
+    // there's at least one new URL (or none) in `currentTokens`
+    Object.keys(currentTokens).forEach((curUrl) => {
+      if (!storeTokens[curUrl]) {
+        // not a stored/known URL: must be new
+        newUrls.push(curUrl);
+      }
+    });
```
bnalisnikovskiy, line 67: As a variant, with less iterations count 
``` 
Object.keys(currentTokens).forEach((url) => {
    if(storeTokens[url]) {
      updatedUrls.push(url); // note that this is not actual updated cloud. Just an existing one. We don't check tockens inside 
    } else {
      newUrls.push(url)
    }
    })
  const oldUrls = Object(_.omit(storeTokens, updatedUrls) || {}).keys
```

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -115,6 +158,7 @@ export const ExtendedCloudProvider = function (props) {
   // autorun calls on any cloudStore update (to often)
   autorun(() => {
     const cloudUrlsToUpdate = [];
```
bnalisnikovskiy, line 160: we don't use this anymore (probably should?)

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -31,57 +31,100 @@ const pr = new ExtendedCloudProviderStore();
...
-      // stop setInterval, just in case and remove EC
-      pr.store.extendedClouds[url].stopUpdateCloudByTimeOut();
-      delete pr.store.extendedClouds[url];
-    });
-  }
+const _updateStore = function (
+  tokens,
+  cloudUrlsToAdd,
+  cloudUrlsToUpdate,
+  cloudsUrlsToDelete
```
bnalisnikovskiy, line 86: Nitpick. Better to use Object as argument if arguments.count > 3. Just to avoid order-bugs

src/renderer/store/ExtendedCloudProvider.js
```diff
@@ -31,57 +31,100 @@ const pr = new ExtendedCloudProviderStore();
...
-    : Object.keys(cloudStore.clouds);
+  cloudsUrlsToDelete.forEach((url) => {
+    // destroy and remove EC
+    pr.store.extendedClouds[url].destroy();
+    delete pr.store.extendedClouds[url];
+  });
 
-  pr.store.tokens = tokens;
-  cloudUrls.forEach((url) => {
+  cloudUrlsToAdd.concat(cloudUrlsToUpdate).forEach((url) => {
```
bnalisnikovskiy, line 96: We always iterate all clouds, even if token still the same.  `cloudUrlsToUpdate` contains all existing clouds (see other comments)

src/store/CloudStore.js
```diff
@@ -95,29 +102,63 @@ export class CloudStore extends Common.Store.ExtensionStore {
...
     extendObservable(this.clouds, { [cloudUrl]: cloud });
   }
+
   /**
+   * Subscribes to a Cloud's change events to know when it gets updated and should
+   *  be written to disk via this store.
+   * @param {Cloud} cloud
+   */
+  listenForChanges(cloud) {
+    Object.values(CLOUD_EVENTS).forEach((eventName) =>
```
bnalisnikovskiy, line 144: Do we need listen `STATUS_CHANGE` on this level?

bnalisnikovskiy (changes requested):
A few comments/questions 
+ store on disk isn't working for me

stefcameron:
> Seems we have a problem with store-on-disk. When I add new Cloud, after 'Synchronize Selected click' I redirected to SyncView and it appears in the table (all is good), but cloud-store.json doesn't have that cloud.

Yes, I'm aware of that. I have the same problem too, and no idea what's going on. I've spent a lot of time scratching my head. I'm hoping someone from the Lens team will help me figure it out before I merge this.

src/common/ExtendedCloud.js
```diff
@@ -425,97 +558,194 @@ export class ExtendedCloud {
...
+
+    // listen to our own event to fetch new data; this allows us to de-duplicate calls
+    //  to fetchData() when the Cloud's properties get updated multiple times in
+    //  succession because of CloudStore updates or expired tokens that got refreshed
+    this.addEventListener(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this.onFetchData);
+
+    // start fetching new data on a regular interval (i.e. don't just rely on
+    //  `cloud` properties changing to trigger a fetch) so we discover changes
+    //  in the Cloud (e.g. new clusters, namespaces removed, etc)
     this._updateInterval = setInterval(() => {
```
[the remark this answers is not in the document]
stefcameron: It's not doing any harm. Whether we need to have the interval or not, the EC still needs to be properly destroyed by AddCloudInstance because the EC, at the least, listens to the Cloud internally.

src/renderer/store/ExtendedCloudProvider.js
stefcameron: Thanks for the suggestion, but I don't think this saves a loop because you're looping once with `Object.keys(currentTokens)` and then another time with `_.omit(...)`. In the worst case scenario, the length of `currentTokens` is equal to that of `storeTokens`.

src/renderer/store/ExtendedCloudProvider.js
stefcameron: Thanks for mentioning this! I see that comparison now, and that piece of code is actually no longer being used. I'm reworking this a bit to clean it up.

src/renderer/store/ExtendedCloudProvider.js
stefcameron: Right! Related to https://github.com/Mirantis/lens-extension-cc/pull/711#discussion_r800496150

src/renderer/store/ExtendedCloudProvider.js
stefcameron: Right! Related to https://github.com/Mirantis/lens-extension-cc/pull/711#discussion_r800496150

src/renderer/store/ExtendedCloudProvider.js
bnalisnikovskiy: Looks good now.
Waiting for store-on-disk fix and after testing I can approve that. Hard to say what can goes wrong right now. I spent almost full day on testing and  checking this feature at first realisation :)

stefcameron:
@bnalisnikovskiy The changes in https://github.com/Mirantis/lens-extension-cc/pull/711/commits/646f667f4e7fa914fa066e1cee0769033d307c95 fix the CloudStore disk issue.

bnalisnikovskiy (approved):
LGTM! CloudStore works well. Can be merged after console.log clean up

PR #717, opened 2022-02-07
Use existing Namespace class in ExtededCloud instead of anonymous object

stefcameron:
We have a Namespace, and a Cluster class, yet `ExtendedCloud.namespaces`
was a list of anonymous objects with properties.

This moves it to being a list of Namespace objects using the existing
Namespace class, adding necessary properties to the class for clusters,
credentials, and SSH keys.

We're currently missing classes for Credentials and SSH keys. We'll need
to add those in a future PR. I will create a ticket for that. We should
propagate the use of raw API objects throughout the code.

This also changes the ExtendedCloud to always fetch ALL namespaces instead
of just the synced ones since we need to know what all exists at all times,
and adds a convenience `syncedNamespaces` getter that filters that list
to just what the Cloud is configured to sync for purposes of UI display
and (future) SyncManager.

### DIFF
--- a/src/renderer/store/Namespace.js
+++ b/src/renderer/store/Namespace.js
@@ -1,4 +1,5 @@
 import * as rtv from 'rtvjs';
+import { Cluster } from './Cluster';
 
 /**
  * MCC project/namespace.
@@ -25,6 +26,10 @@ export class Namespace {
         }
       );
 
+    let _clusters = null;
+    let _sshKeys = null;
+    let _credentials = null;
+
     /** @member {string} */
     this.id = data.metadata.uid;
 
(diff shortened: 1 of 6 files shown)


stefcameron:
__NOTE:__ As on #711, any code followed by a `// DEBUG` comment will be removed before I merge this. I just left those statements in for now, in case I have to make larger changes after review.

src/renderer/store/Namespace.js
```diff
@@ -36,5 +41,96 @@ export class Namespace {
...
+   * @member {number} clusterCount Number of clusters in this namespace.
+   */
+  get clusterCount() {
+    return this.clusters?.length || 0;
+  }
+
+  /**
+   * @member {number} clusterCount Number of SSH keys in this namespace.
+   */
+  get sshKeyCount() {
```
bnalisnikovskiy, line 133: :+1:

bnalisnikovskiy (approved):
Looks good!

PR #719, opened 2022-02-08
[PRODX-21659]: fixed checkbox icon font size to make it integer

AntonNos:
Screenshot 2022-02-08 at 13 21 12


stefcameron (approved):
Much better, thanks!

PR #720, opened 2022-02-08
[PRODX-21652]: added open in browser logic

AntonNos:


src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -99,7 +99,9 @@ const cloudMenuItems = [
   {
     title: `(WIP) ${contextMenus.cloud.openInBrowser()}`,
     name: 'openInBrowser',
-    onClick: () => {},
+    onClick: (url) => {
+      require('electron').shell.openExternal(url);
```
bnalisnikovskiy, line 103: use Common.Util for that

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -205,7 +207,10 @@ export const EnhancedTableRow = ({ extendedCloud }) => {
             <MenuActions onOpen={onOpen}>
               {cloudMenuItems.map((item) => {
                 return (
-                  <MenuItem key={`cloud-${item.name}`} onClick={item.onClick}>
+                  <MenuItem
+                    key={`cloud-${item.name}`}
+                    onClick={() => item.onClick(extendedCloud.cloud.cloudUrl)}
```
bnalisnikovskiy, line 212: It's a not a good variant, we don't need urls for other items. Better to pass full extendedCloud and take what we need in every particular case. Or figure out something better

bnalisnikovskiy (changes requested):

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -99,7 +99,9 @@ const cloudMenuItems = [
   {
     title: `(WIP) ${contextMenus.cloud.openInBrowser()}`,
```
bnalisnikovskiy, line 100: Also, if feature is ready 'WIP' label has to be removed

src/renderer/components/EnhancedTable/EnhancedTableRow.js
stefcameron: Please use the openBrowser() function in our netUtil.js module as it has an additional security check.

src/renderer/components/EnhancedTable/EnhancedTableRow.js
stefcameron: Yes, I think it would be better to have each item expect an EC and then do what it needs to do with it.

stefcameron (changes requested):
See comments

bnalisnikovskiy (approved):
Looks good. 
One moment. I think we can make that more efficient:
```
const cloudMenuItems = getCloudMenuItems(extendedCloud)
const getCloudMenuItems = (extendedCloud) => [
...
{
    title: `(WIP) ${contextMenus.cloud. openInBrowser()}`,
    name: 'openInBrowser',
    onClick: () =>  openBrowser(extendedCloud.cloud.cloudUrl),
  },
  ...
]
```
Then in component we don't care about functions 
```
<MenuItem key={`cloud-${item.name}`} onClick={item.onClick}>

```

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -97,9 +98,9 @@ const cloudMenuItems = [
     onClick: () => {},
   },
   {
-    title: `(WIP) ${contextMenus.cloud.openInBrowser()}`,
+    title: `${contextMenus.cloud.openInBrowser()}`,
     name: 'openInBrowser',
-    onClick: () => {},
+    onClick: () => openBrowser(extendedCloud.cloud.cloudUrl),
```
stefcameron, line 103: I believe the ask from Bogdan and I was that each `onClick` gets an ExtendedCloud object as a parameter. That seems more flexible in terms of design than assuming that each item will act on the same EC and each item getting the EC from context instead of the EC being given to it.

stefcameron: Personally, I think we achieve better separation of concerns if we do it this way (like you had it in [this commit](https://github.com/Mirantis/lens-extension-cc/pull/720/commits/4f22c36d5849c87766e3d67ba66250b9b0670583)) instead of going for efficiency. I don't think it's really more efficient one way or another, and more a matter of not having a click handler that has already pre-determined what EC it's going to operate on.

AntonNos: Reverted to previous commit

PR #721, opened 2022-02-08
PRODX-21649: Added reconnect logic

bnalisnikovskiy:
Fire `extCloud.cloud.connect()` on 'reconnect' click 

image

### DIFF
--- a/src/renderer/components/EnhancedTable/EnhancedTableRow.js
+++ b/src/renderer/components/EnhancedTable/EnhancedTableRow.js
@@ -81,7 +81,7 @@ const getStatus = (isCloudConnected) => {
       };
 };
 
-const getCloudMenuItems = (extendedCloud) =>  [
+const getCloudMenuItems = (extendedCloud) => [
   {
     title: contextMenus.cloud.reconnect(),
     name: 'reconnect',
--- a/src/renderer/components/EnhancedTable/EnhancedTableRow.js
+++ b/src/renderer/components/EnhancedTable/EnhancedTableRow.js
@@ -81,11 +81,11 @@ const getStatus = (isCloudConnected) => {
       };
 };
 
-const getCloudMenuItems = (extendedCloud) => [
+const cloudMenuItems = [
   {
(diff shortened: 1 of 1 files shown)


src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -80,11 +80,11 @@ const getStatus = (isCloudConnected) => {
...
 };
 
-const cloudMenuItems = [
+const getCloudMenuItems = (extendedCloud) => [
   {
-    title: `(WIP) ${contextMenus.cloud.reconnect()}`,
+    title: contextMenus.cloud.reconnect(),
     name: 'reconnect',
-    onClick: () => {},
+    onClick: () => extendedCloud.cloud.connect(),
```
stefcameron, line 87: On #720, we went back to passing the EC through the `onClick()` handler. That was my request. I see that as more "separation of concerns" than having handlers already bound to a pre-determined object that isn't passed-in.

I realize in this case all menu items will act on the same EC, so it seems more efficient to just pass it in once via `getCloudMenuItems()`, but I don't think it's really more "efficient". More _convenient_, perhaps, but passing-in the EC seems to me like it makes the event handler more portable and predictable.

All that to say, there may be a merge conflict, or maybe this would slip through, so just in case, please update to match #720 so we're consistent, before you merge.

stefcameron (approved):
This is great! Will be finally more convenient than quitting Lens, editing cloud-store.json, and restarting Lens in order to get a Cloud working again... 😉

src/renderer/components/EnhancedTable/EnhancedTableRow.js
bnalisnikovskiy: Ok, I'll change this. 
Btw, we can do ` const cloudMenuItems = useMemo(() => getCloudMenuItems(extendedCloud), [extendedCloud])` to optimise first approach. Then is should be really more efficient
Anyway, if we find that eventHandler needs something more than only EC, we can transform it to `getCloudMenuItems`.

PR #722, opened 2022-02-09
Ensure ExtendedCloud.namespaces is always an array

stefcameron:
We were using the `namespaces` property with a double purpose:
- if it was null, then this indicated we hadn't loaded successfully
  for the first time yet
- if it was an array, then we had loaded, and we had data

This leads to "protective" code like `extCloud.namespaces || []` in
many places.

This change makes `namespaces` be always an array, and empty if we've
never loaded yet successfully at least once.

It changes `ExtendedCloud.loading` to `ExtendedCloud.loaded` which is
false until we've completed one successful data load. After that, it's
_always_ true. `ExtendedCloud.fetching` keeps flipping from true
when we're fetching, back to false when we aren't.

The `error` property also gets its own event, and AddCloudInstance
is made to listen for `ERROR_CHANGE` in case there's an error during
the initial load, since it only cares about the intial load.

### DIFF
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -14,22 +14,15 @@ export const EXTENDED_CLOUD_EVENTS = Object.freeze({
    *
    * Expected signature: `(extCloud: ExtendedCloud) => void`
    */
-  LOADING_CHANGE: 'loadingChange',
+  LOADED: 'loaded',
 
   /**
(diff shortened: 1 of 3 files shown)


bnalisnikovskiy (approved):
LGTM
Not tested, though
If some bug appears, it will be fixed in separate ticket.

PR #725, opened 2022-02-09
PRODX-21692 Added more Classes

bnalisnikovskiy:
Added  next classes: 
- `SshKey`
- `Proxy`
- `License`
- `Credential`
Updated `Cluster` and `Namespace` classes according to [the ticket description  ](https://mirantis.jira.com/browse/PRODX-21692)
Also instead of `namespace` name (string type) from API call, we store `Namespace` in those Classes mentioned above

### DIFF
--- a/src/renderer/store/Credential.js
+++ b/src/renderer/store/Credential.js
@@ -1,0 +1,71 @@
+import * as rtv from 'rtvjs';
+
+const openStackCredentialSpec = {
+  apiVersion: rtv.STRING,
+  kind: [rtv.REQUIRED, rtv.STRING],
+  metadata: {
+    name: rtv.STRING,
+    namespace: rtv.STRING,
+    uid: rtv.STRING,
+    creationTimestamp: rtv.STRING, // ISO8601 timestamp
+    finalizers: [rtv.ARRAY, rtv.OBJECT],
+    managedFields: [rtv.ARRAY, { $: [rtv.OBJECT] }], // complex nested object
+    resourceVersion: rtv.STRING,
+    labels: [
+      rtv.OPTIONAL,
+      {
+        'kaas.mirantis.com/provider': [rtv.OPTIONAL, rtv.STRING],
+        'kaas.mirantis.com/region': [rtv.OPTIONAL, rtv.STRING],
+      },
+    ],
+  },
+  spec: {
+    regionName: rtv.STRING,
+    auth: {
+      authURL: rtv.STRING,
+      password: rtv.OBJECT, // do we need details here?
+      projectID: rtv.STRING,
+      userDomainName: rtv.STRING,
+      userName: rtv.STRING,
+    },
+  },
+  status: {
+    valid: rtv.BOOLEAN,
+  },
+};
+
+export class Credential {
+  constructor(data, namespace) {
+    DEV_ENV &&
+      rtv.verify(
+        { data, namespace },
+        {
+          namespace: [rtv.REQUIRED, rtv.STRING],
+          data: {
+            openstackcredential: [
+              rtv.OPTIONAL,
+              rtv.ARRAY,
+              { $: [openStackCredentialSpec] },
+            ],
+            awscredential: [rtv.OPTIONAL, rtv.ARRAY],
+            byocredential: [rtv.OPTIONAL, rtv.ARRAY],
+          },
+        }
+      );
+    this.namespace = namespace;
+    this.openstackcredential = data?.openstackcredential || [];
+    this.awscredential = data?.awscredential || [];
+    this.byocredential = data?.byocredential || [];
+
+    Object.defineProperty(this, 'allCredentialsCount', {
+      enumerable: true,
+      get() {
+        return (
+          this.openstackcredential.length +
+          this.awscredential.length +
+          this.byocredential.length
+        );
+      },
+    });
+  }
+}
--- a/src/renderer/store/Credential.js
+++ b/src/renderer/store/Credential.js
@@ -15,7 +15,7 @@ const openStackCredentialSpec = {
     name: rtv.STRING,
     namespace: rtv.STRING,
     uid: rtv.STRING,
-    finalizers: [rtv.ARRAY, rtv.OBJECT],
+    finalizers: [rtv.OPTIONAL, rtv.ARRAY, rtv.OBJECT],
     managedFields: [rtv.ARRAY, { $: [rtv.OBJECT] }], // complex nested object
     resourceVersion: rtv.STRING,
     labels: [
@@ -49,10 +49,11 @@ export class Credential extends ApiObject {
 
     /** @member {string} */
     this.kind = data.kind;
-    /** @member {string} */
 
+    /** @member {string}  namespaceName */
     this.namespace = data.metadata.namespace;
 
+    /** @member {string} */
     this.region = get(
       data.metadata,
       'labels["kaas.mirantis.com/region"]',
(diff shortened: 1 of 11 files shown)


stefcameron:
I noticed this PR, but I believe you posted it before I updated the info in the ticket, so I'll wait for you to update the code.

src/renderer/store/Credential.js
```diff
@@ -0,0 +1,82 @@
...
+    this.provider = get(
+      data.metadata,
+      'labels["kaas.mirantis.com/provider"]',
+      null
+    );
+
+    // TODO this need to find and  define
+    this.status = data.status.valid ? 'Connected' : 'Disconnected';
+    this.source = null;
+    this.labels = [];
```
bnalisnikovskiy, line 80: FYI @stefcameron 
Regarding to design we need next fields: 
- Status
- Source ('MCC')
- labels 
Currently, I don't know where we can take it. This has to be updated for each class. 
image

stefcameron: These three properties are "presentation" properties related to the Catalog Entity that will use a class instance such as `SshKey` or `Cluster` for its data.

So in my view, `source` (which will always be the string "MCC"), `status`, and `labels` belong in the related entity (for an `SshKey` instance, that will be the `SshKeyEntity`, and as an example, `status` is determined here: https://github.com/Mirantis/lens-extension-cc/blob/master/src/catalog/SshKeyEntity.ts#L19).

For the labels, we know what they will be based on each entity type, and when we [add entities to the Catalog](https://github.com/Mirantis/lens-extension-cc/blob/master/src/main/IpcMain.js#L238), we will generate labels for them based on the underlying class' properties. I'm not doing that for the fake items, but you can see that the generic [entity model metadata](https://github.com/Mirantis/lens-extension-cc/blob/master/src/typesets.js#L18) defines a `labels` property, so every Catalog entity has labels.

For example, a [CredentialEntity](https://github.com/Mirantis/lens-extension-cc/blob/master/src/catalog/CredentialEntity.ts) will get 2 labels: "project" and "managementCluster". The `SyncManager` -- who will be the thing that creates and adds these entities to the Catalog -- should be able to look at a `Credential` class instance and determine what it's namespace (i.e. "project") is, as well as what it's mgmt cluster (i.e. the `ExtendedCloud`) to which it belongs (and then it's a matter of `credential.extendedCloud.cloud.name` for the mgmt cluster label's value).

Also, looking into this made me remember that I had already figured out the basis of the models for most entity types here:

- https://github.com/Mirantis/lens-extension-cc/blob/master/src/typesets.js#L55
- https://github.com/Mirantis/lens-extension-cc/blob/master/src/typesets.js#L81
- https://github.com/Mirantis/lens-extension-cc/blob/master/src/typesets.js#L103
- https://github.com/Mirantis/lens-extension-cc/blob/master/src/typesets.js#L128

Just missing License. I'm sorry I forgot to mention this earlier. I did the Catalog entity work back in December before I went on Christmas vacation.

src/renderer/store/Credential.js
```diff
@@ -0,0 +1,82 @@
...
+      userName: rtv.STRING,
+    },
+  },
+  status: {
+    valid: rtv.BOOLEAN,
+  },
+};
+
+export class Credential extends ApiObject {
+  constructor(data, namespace) {
```
stefcameron, line 46: The constructor should take another parameter that is the credential type, which should be one of the `credentialTypes` values.

src/renderer/store/Namespace.js
```diff
@@ -78,10 +84,9 @@ export class Namespace {
...
       },
       set(newValue) {
-        // TODO: update the shape to check for an SshKey class instance
         rtv.verify(
           { sshKeys: newValue },
-          { sshKeys: [rtv.EXPECTED, rtv.ARRAY, { $: [rtv.OBJECT] }] }
+          { sshKeys: [rtv.EXPECTED, rtv.ARRAY, { ctor: SshKey }] }
         );
         if (newValue !== _sshKeys) {
           _sshKeys = newValue || null;
```
stefcameron, line 92: Default to `[]`

src/renderer/store/Namespace.js
```diff
@@ -118,6 +126,56 @@ export class Namespace {
         }
```
stefcameron, line 126: Default to `[]`

src/renderer/store/Namespace.js
```diff
@@ -118,6 +126,56 @@ export class Namespace {
...
+          }
+        );
+        if (newValue !== _proxies) {
+          _proxies = newValue || null;
+        }
+      },
+    });
+
+    /**
+     * @member {Array<Object>|null} proxies Proxies in this namespace. `null` if unknown.
```
stefcameron, line 156: "proxies" -> "licenses"

stefcameron: Should also be just `{Array<License>}` for the type.

src/renderer/store/Namespace.js
```diff
@@ -29,6 +33,8 @@ export class Namespace {
     let _clusters = null;
     let _sshKeys = null;
     let _credentials = null;
+    let _proxies = null;
+    let _licenses = null;
```
stefcameron, line 37: I think it'll be better for downstream code if we ensure that these properties are always arrays, and empty if none/unknown. If we don't know what they are, that's because we haven't fetched the namespace, and we won't have a Namespace class instance in that case anyway.

```suggestion
    let _clusters = [];
    let _sshKeys = [];
    let _credentials = [];
    let _proxies = [];
    let _licenses = [];
```

src/renderer/store/SshKey.js
```diff
@@ -0,0 +1,44 @@
...
+    /** @member {string} */
+    this.kind = data.kind;
+
+    /** @member {string} */
+    this.publicKey = data.spec.publicKey;
+
+    // TODO this need to find and  define
+    this.status = null;
+    this.source = null;
+    this.labels = [];
```
stefcameron, line 42: Can be removed.

stefcameron (changes requested):
Just minor changes. After we merge this, I'll do a second pass to reconcile gaps I've now identified between this work (that we need) and the initial discovery work I did for catalog entities in https://github.com/Mirantis/lens-extension-cc/tree/master/src/catalog

stefcameron (approved):
Thanks for those updates!

Looks good. Just maybe we should rethink the `Namespace.credentials` property. See my comment and let me know what you think about it.

src/renderer/store/Namespace.js
```diff
@@ -89,14 +87,17 @@ export class Namespace {
...
     });
 
     /**
-     * @member {Array<Object>|null} credentials Credentials in this namespace. `null` if unknown.
-     *  Empty if none.
+     * @member {Object} credentials grouped by type in this namespace. Empty object by default.
+     * @param {Array<Credential>} credentials.awscredential
+     * @param {Array<Credential>} credentials.byocredential
+     * @param {Array<Credential>} credentials.openstackcredential
+     * @param {number} credentials.allCredentialsCount
```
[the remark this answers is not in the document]
bnalisnikovskiy: Thanks for your comment. I'll improve JSDoc as you said. As I can see, Webstorm doesn't handle them so well as VSCode does

About credentials structure, it sounds good and simple structure looks preferred. Only one thing I worried about is body response. We do three different API calls for them, so maybe `response.body` has some major diversities, it's hard to predict without real data/description (what to expect). 

I event don't know what is better on current stage, keep them as it is for now, or change to single mixed-types array hoping that data signature the same...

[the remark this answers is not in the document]
bnalisnikovskiy: Updated

stefcameron (approved):
Looks good!

src/renderer/store/Credential.js
stefcameron: Addressed in #743 and #744

PR #726, opened 2022-02-09
Improve reconnect logic

bnalisnikovskiy:
During 'reconnect' => connect to Cloud and fetch data for ExtendedCloud


src/common/ExtendedCloud.js
```diff
@@ -672,6 +672,14 @@ export class ExtendedCloud extends EventDispatcher {
     this.fetching = false;
   }
 
+  /**
+   * @desc reconnect cloud and fetch data for EC
+   * @return {Promise<void>}
+   */
+  reconnect = async () => {
+    await this.cloud.connect();
+    this.dispatchEvent(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this);
```
stefcameron, line 681: Actually, it's more complicated than this as I look at the code. `Cloud.connect()` is deceiving because it's async and makes you think you can just `await` and when it comes back, you're done and connected, but what actually happens is, while it awaits internally on loading the config, it then has to open the browser and it just returns. Then we wait for the user to interact in the browser and trigger the SSO flow to make the request back to Lens.

So what we really need to do here is `await this.cloud.connect()` and then start listening for the Cloud's `STATUS_CHANGE` event. In the handler, if `this.cloud.isConnected()` is true, then dispatch the FETCH_DATA event. Either way, remove the STATUS_CHANGE listener in the handler since the EC doesn't need to always listen for status changes.

stefcameron (changes requested):
Turns out reconnecting is a bit more complicated than I originally thought...

src/common/ExtendedCloud.js
stefcameron: Actually, EC should probably check the `Cloud.status` property instead of `Cloud.isConnected()`.

stefcameron: Also would be nice to add an empty line after this method. 🙂

stefcameron: While working on #727, I wanted to test the reconnection scenario with and without connection errors, so I ended-up implementing `ExtendedCloud.reconnect()` locally, but didn't include it in my PR since you're doing that here and adding that feature wasn't my goal in my PR.

Here's the code I wrote and tested:

```javascript
  async reconnect() {
    // WARNING: does not block because goes to browser and waits for user
    await this.cloud.connect();

    if (this.cloud.connecting) {
      const handler = () => {
        if (this.cloud.connected) {
          this.cloud.removeEventListener(CLOUD_EVENTS.STATUS_CHANGE, handler);
          this.dispatchEvent(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this);
        }
      };
      this.cloud.addEventListener(CLOUD_EVENTS.STATUS_CHANGE, handler);
    } else {
      this.error = this.cloud.connectError;
    }
  }
```

bnalisnikovskiy: Updated and works fine now. Thanks!

stefcameron (approved):
Looks good! Just one little request before you merge... 🙂

src/common/ExtendedCloud.js
```diff
@@ -700,6 +700,26 @@ export class ExtendedCloud extends EventDispatcher {
...
+          this.cloud.removeEventListener(CLOUD_EVENTS.STATUS_CHANGE, handler);
+          this.dispatchEvent(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this);
+        }
+      };
+      this.cloud.addEventListener(CLOUD_EVENTS.STATUS_CHANGE, handler);
+    } else {
+      this.error = this.cloud.connectError;
+    }
+  }
   /** @returns {string} String representation of this ExtendedCloud for logging/debugging. */
```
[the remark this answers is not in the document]
bnalisnikovskiy: Will do. Is it possible to add this as Prettier/Eslint rule? Usually, I delegate this work to them

PR #727, opened 2022-02-09
Consolidate Cloud.isConnected() and Cloud.status, fix connected notion

stefcameron:
I noticed that we ended-up with 2 ways of determining if a Cloud is
"connected" (which will lead to bugs down the line), and also an
incorrect notion of what "connected" means.

Now, we have a single way, `Cloud.status`, and a helper with
`Cloud.connected` (a boolean getter instead of a function, which
is true if status === CONNECTED, false otherwise, for cases where
the binary state is all we care about). Both properties are now
derived from the same set of logic in the `Cloud.status` getter,
so they won't diverge.

Also, "connected" really means the Cloud has all the necessary info
in order to make an API request. Without the `config` loaded, we're
technically not connected yet. So now, if we have everything except
the config, we consider this "connecting" because we have tokens,
but we just haven't gotten the config yet.

This covers the case where the Cloud is restored from disk with valid
tokens as the Cloud doesn't auto-fetch its config. The ExtendedCloud
will fetch the config when it does its initial data fetch. Then the
Cloud will become fully "connected".

I also tested some error scenarios in `Cloud.connect()` and ended-up
tweaking how `ExtendedCloud.fetchData()` reacts as a result. I think
things are more robust now than they were before.

### Preview

Here's what it looks like now when opening Lens and we already have a Cloud with valid tokens and the ExtendedCloud does its initial data fetch automatically:

lex-connecting-status

And here's what it looks like, in terms of status, when we load Lens and we have a disconnected Cloud, and we "reconnect" to it:

lex-reconnect-with-status

### DIFF
--- a/src/common/Cloud.js
+++ b/src/common/Cloud.js
@@ -13,11 +13,11 @@ import { EventDispatcher } from './EventDispatcher';
 
 /**
  * Determines if a date has passed.
- * @param {Date} date
- * @returns {boolean} true if the date has passed; false otherwise.
+ * @param {Date} [date]
+ * @returns {boolean} True if the date has passed or is not a `Date`; false otherwise.
  */
 const hasPassed = function (date) {
-  return date.getTime() - Date.now() < 0;
+  return date instanceof Date ? date.getTime() - Date.now() < 0 : false;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


bnalisnikovskiy (approved):
Looks good! :+1:

PR #728, opened 2022-02-09
[PRODX-21648]: added remove cloud logic

AntonNos:
Screenshot 2022-02-14 at 22 18 27
Screenshot 2022-02-14 at 22 19 54

### DIFF
--- a/src/renderer/components/EnhancedTable/EnhancedTableRow.js
+++ b/src/renderer/components/EnhancedTable/EnhancedTableRow.js
@@ -151,7 +151,7 @@ const cloudMenuItems = [
           ),
         });
       } else {
-        if (extendedCloud.cloud.syncNamespaces > 0) {
+        if (extendedCloud.namespaces.length > 0) {
           ConfirmDialog.open({
             ok: () => {
               cloudStore.removeCloud(extendedCloud.cloud.cloudUrl);
@@ -162,7 +162,7 @@ const cloudMenuItems = [
                 dangerouslySetInnerHTML={{
                   __html: contextMenus.cloud.confirmDialog.messageHtml(
                     extendedCloud.cloud.name,
-                    extendedCloud.cloud.syncNamespaces
+                    extendedCloud.namespaces
                   ),
                 }}
               />
--- a/src/renderer/components/EnhancedTable/EnhancedTableRow.js
+++ b/src/renderer/components/EnhancedTable/EnhancedTableRow.js
@@ -123,7 +123,7 @@ const cloudMenuItems = [
     onClick: (extendedCloud) => extendedCloud.reconnect(),
   },
   {
-    title: `${contextMenus.cloud.remove()}`,
+    title: contextMenus.cloud.remove(),
     name: 'remove',
     onClick: (extendedCloud) => {
       ConfirmDialog.open({
--- a/src/renderer/components/EnhancedTable/EnhancedTableRow.js
+++ b/src/renderer/components/EnhancedTable/EnhancedTableRow.js
@@ -126,8 +126,11 @@ const cloudMenuItems = [
     title: contextMenus.cloud.remove(),
     name: 'remove',
     onClick: (extendedCloud) => {
-      console.log('test ', extendedCloud.cloud.syncAll);
-      if (!extendedCloud.cloud.syncAll && extendedCloud.cloud.syncNamespaces.length === 0) {
+      if (
+        !extendedCloud.cloud.connected ||
+        (!extendedCloud.cloud.syncAll &&
+          extendedCloud.cloud.syncNamespaces.length === 0)
+      ) {
         cloudStore.removeCloud(extendedCloud.cloud.cloudUrl);
       } else {
         ConfirmDialog.open({
@@ -138,9 +141,8 @@ const cloudMenuItems = [
           message: (
             <div
               dangerouslySetInnerHTML={{
-                __html: contextMenus.cloud.confirmDialog.messageHtml(
-                  extendedCloud
-                ),
+                __html:
+                  contextMenus.cloud.confirmDialog.messageHtml(extendedCloud),
               }}
             />
           ),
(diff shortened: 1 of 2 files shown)


src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -88,9 +89,26 @@ const cloudMenuItems = [
...
   },
   {
-    title: `(WIP) ${contextMenus.cloud.remove()}`,
+    title: `${contextMenus.cloud.remove()}`,
     name: 'remove',
-    onClick: () => {},
+    onClick: (extendedCloud) => {
+      ConfirmDialog.open({
+        ok: () => {
+          extendedCloud.destroy();
```
stefcameron, line 97: We shouldn't call this directly. Removing it from the CloudStore will result in the EC being automatically destroyed by the ExtendedCloudProvider.

src/strings.ts
```diff
@@ -333,6 +333,11 @@ export const contextMenus: Dict = {
     remove: () => 'Remove',
     sync: () => 'Sync now',
     openInBrowser: () => 'Open in browser',
+    confirmDialog: {
+      messageHtml: (cloudName) =>
+        `Are you sure you want to remove the <b>${cloudName}</b> cloud from Lens?`,
```
stefcameron, line 338:
```suggestion
        `Are you sure you want to remove the <b>${cloudName}</b> Management Cluster from Lens?`,
```

stefcameron (changes requested):
Finally, the remove option! That will help with local testing... 😄

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -88,9 +89,25 @@ const cloudMenuItems = [
     onClick: () => {},
   },
   {
-    title: `(WIP) ${contextMenus.cloud.remove()}`,
+    title: `${contextMenus.cloud.remove()}`,
```
bnalisnikovskiy, line 92: NIt: Just `contextMenus.cloud.remove()` without ``${}``

bnalisnikovskiy (approved):
Looks good. One nitpick

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -88,9 +89,25 @@ const cloudMenuItems = [
...
+        },
+        labelOk: contextMenus.cloud.confirmDialog.confirmButtonLabel(),
+        message: (
+          <p
+            dangerouslySetInnerHTML={{
+              __html: contextMenus.cloud.confirmDialog.messageHtml(
+                extendedCloud.cloud.name
+              ),
+            }}
+          />
```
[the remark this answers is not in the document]
AntonNos: There is one more question - what should we do if management cluster is `disconnected`? For now, I removing it without any dialog message. Is it okay, or should we fetch data and also show the following projects and their associated catalog items?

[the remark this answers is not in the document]
stefcameron: Good question. If the Cloud is disconnected, we may (likely) have connected to it at least once and synced entities into Lens, so these entities will still get removed. But if Lens starts with the Cloud disconnected, we won't know what all the projects are -- if they're syncing all projects. If they're syncing specific projects, we can get the list from `extendedCloud.cloud.syncedNamespaces`.

So if `!cloud.connected` OR `!extendedCloud.loaded`, then
- if `!cloud.syncAll`, use `cloud.syncedNamespaces` to list the projects in the confirmation dialog (this list must have at least one name in it)
- if `cloud.syncAll`, use the same message, except don't show the list of projects (and replace "items:" with "items." -- a period instead of a colon) because we don't know what they are.

Else (which means `cloud.connected` AND `extendedCloud.loaded`), then `extendedCloud.syncedNamespaces` should be the list of projects whose items will be removed:
- if `syncedNamespaces.length > 0`, show the message with the project names in bullet points.
- if `syncedNamespaces.length <= 0`, don't show the confirmation dialog at all; there are no namespaces, so no items in the Catalog related to them, so nothing will be lost if the Cloud is removed.

I think that should cover all the cases. Can you think of any others?

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -122,9 +123,56 @@ const cloudMenuItems = [
...
+                __html: contextMenus.cloud.confirmDialog.messageHtml(
+                  confirmDialogMessageParams.name,
+                  confirmDialogMessageParams.projects
+                ),
+              }}
+            />
+          ),
+        });
+      } else {
+        if (extendedCloud.cloud.syncNamespaces > 0) {
```
stefcameron, line 154: You have a bug here, missing `.length` on the array.

BUT: It might be safer in the long run to also check that we aren't syncing all namespaces. But then to go further, given that in this `else` block, we're both connected AND loaded, the even-safer check would probably be `if (extendedCloud.namespaces.length > 0)` because given we're connected and loaded, if that list is empty, we can be sure that there are no namespaces _being_ synced (whether `Cloud.syncAll` is true or `Cloud.syncNamespaces.length > 0`).

src/strings.ts
```diff
@@ -339,6 +339,24 @@ export const contextMenus: Dict = {
...
+      messageHtml: (extCloudName, extCloudProjects) =>
+        `
+        <p>Removing management cluster “${extCloudName}” will also remove the following projects and their associated catalog items${
+          extCloudProjects ? ':' : '.'
+        }</p>
+        ${
+          extCloudProjects.length > 0 &&
+          `<ul style="padding: 12px;">
+            ${extCloudProjects
+              .map((project) => `<li>- ${project}</li>`)
```
stefcameron, line 352: We don't need the hyphen here; the `<li>` will provide a bullet already.

Could you post a screenshot of this dialog when there's a list and when there isn't?

AntonNos: Screenshots was added to description.

src/renderer/components/EnhancedTable/EnhancedTableRow.js
AntonNos: Yes, thanks for letting me know, that I missed `.length`.

stefcameron (approved):
This will work! Just one last thing to tweak before you merge:

image

Can you add an empty line between the two paragraphs in this case?

PR #730, opened 2022-02-10
[PRODX-21534]: Added spinner to AddCloudInstance component

AntonNos:
Screenshot 2022-02-10 at 15 33 51
Screenshot 2022-02-11 at 14 38 25

### DIFF
--- a/src/renderer/components/GlobalPage/ConnectionBlock.js
+++ b/src/renderer/components/GlobalPage/ConnectionBlock.js
@@ -60,7 +60,7 @@ const nameAlreadyUsed = {
     !Object.values(cloudStore.clouds).find(({ name }) => name === value),
 };
 
-export const ConnectionBlock = ({ extCloudLoading, handleClusterConnect }) => {
+export const ConnectionBlock = ({ loading, handleClusterConnect }) => {
   const [clusterName, setClusterName] = useState('');
   const [clusterUrl, setClusterUrl] = useState('');
   const [showInfoBox, setShowInfoBox] = useState(true);
@@ -72,17 +72,9 @@ export const ConnectionBlock = ({ extCloudLoading, handleClusterConnect }) => {
     setClusterUrl(value);
   };
 
-  const handleConnectClick = (
-    managementClusterUrl,
-    setManagementClusterUrl,
-    managementClusterName
-  ) => {
+  const handleConnectClick = () => {
     setShowInfoBox(false);
-    handleClusterConnect(
-      managementClusterUrl,
-      setManagementClusterUrl,
-      managementClusterName
-    );
+    handleClusterConnect(clusterUrl, clusterName);
   };
 
   return (
@@ -100,7 +92,7 @@ export const ConnectionBlock = ({ extCloudLoading, handleClusterConnect }) => {
           id="lecc-cluster-name"
           value={clusterName}
           onChange={setClusterName}
-          disabled={extCloudLoading}
+          disabled={loading}
           validators={[nameSymbols, nameAlreadyUsed]}
           required
           trim
@@ -118,7 +110,7 @@ export const ConnectionBlock = ({ extCloudLoading, handleClusterConnect }) => {
           value={clusterUrl}
           onChange={setUrl}
           validators={[cloudAlreadySynced]}
-          disabled={extCloudLoading}
+          disabled={loading}
           required
           trim
         />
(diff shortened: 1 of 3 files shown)


stefcameron (changes requested):
This is good, but the spinner isn't covering the full span of the async workflow. As soon as the user clicks on the Connect button, and the button gets disabled, the spinner should be visible right up until the moment where we display the list of projects.

AntonNos:
> While we're at it, please also test the error case (i.e. what happens if `Cloud.connect()` or `ExtendedCloud.fetchData()` fails. We need to make sure we have good error handling and show an error message to the user. Probably something like, "An error occurred while connecting to the management cluster and retrieving its projects. Click on the Connect button to try again."

Thank you for suggesting that need to add an error message. I attached screenshot how it looks.

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -65,58 +60,31 @@ const nameAlreadyUsed = {
...
     if (!showInfoBox) {
       setShowInfoBox(true);
     }
     setClusterUrl(value);
   };
 
+  const handleConnectClick = (
+    managementClusterUrl,
+    setManagementClusterUrl,
+    managementClusterName
```
bnalisnikovskiy, line 78: we don't need this arguments. You can use them directly 
```
const handleConnectClick = () => {
    setShowInfoBox(false);
     handleClusterConnect(clusterUrl, clusterName);
  };
  ```

File: src/renderer/components/GlobalPage/ConnectionBlock.js [hunk 2]
```diff
@@ -150,18 +118,19 @@ export const ConnectionBlock = ({
           value={clusterUrl}
           onChange={setUrl}
           validators={[cloudAlreadySynced]}
-          disabled={loading || extCloudLoading}
+          disabled={extCloudLoading}
```
bnalisnikovskiy, line 121: Now it's not only `extCloudLoading`, so lets rename it to `loading`

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -150,18 +118,19 @@ export const ConnectionBlock = ({
...
         <ButtonWrapper>
           <Button
             primary
             type="submit"
-            waiting={loading || extCloudLoading}
             label={connectionBlock.button.label()}
-            onClick={handleConnectClick}
-            disabled={loading || extCloudLoading}
+            onClick={() =>
+              handleConnectClick(clusterUrl, setClusterUrl, clusterName)
```
bnalisnikovskiy, line 131: we don't need arguments. See comment above

bnalisnikovskiy (changes requested):

src/renderer/components/GlobalPage/ConnectionBlock.js
```diff
@@ -150,18 +110,17 @@ export const ConnectionBlock = ({
...
         />
         <ButtonWrapper>
           <Button
             primary
             type="submit"
-            waiting={loading || extCloudLoading}
             label={connectionBlock.button.label()}
-            onClick={handleConnectClick}
-            disabled={loading || extCloudLoading}
+            onClick={() => handleConnectClick()}
```
bnalisnikovskiy, line 122: just `onClick={handleConnectClick}` in this case

bnalisnikovskiy (approved):
Looks good. One comment

stefcameron:
> > While we're at it, please also test the error case (i.e. what happens if `Cloud.connect()` or `ExtendedCloud.fetchData()` fails. We need to make sure we have good error handling and show an error message to the user. Probably something like, "An error occurred while connecting to the management cluster and retrieving its projects. Click on the Connect button to try again."
> 
> Thank you for suggesting that need to add an error message. I attached screenshot how it looks.

And thanks for adding! But I think we should make it an error notification instead of putting the error message in the UI. So the spinner should disappear, and the error notification should appear (Lens will show it in the top/right corner of the window). See https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer/renderer.tsx#L169 for an example.

AntonNos:
> > > While we're at it, please also test the error case (i.e. what happens if `Cloud.connect()` or `ExtendedCloud.fetchData()` fails. We need to make sure we have good error handling and show an error message to the user. Probably something like, "An error occurred while connecting to the management cluster and retrieving its projects. Click on the Connect button to try again."
> > 
> > 
> > Thank you for suggesting that need to add an error message. I attached screenshot how it looks.
> 
> And thanks for adding it! But I think we should make it an error notification instead of putting the error message in the UI. So the spinner should disappear, and the error notification should appear (Lens will show it in the top/right corner of the window). See https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer/renderer.tsx#L169 for an example.

Added error message to Notification instead of UI block

stefcameron (approved):
Thanks for changing it to an error notification!

PR #731, opened 2022-02-10
Add lint rule for class member styling

stefcameron:
So far, Prettier doesn't style class members in any specific way,
so this won't conflict with Prettier styling.


stefcameron:
Let's give it a try and see how we like it.

PR #732, opened 2022-02-14
PRODX-21766 Added Selective sync logic

bnalisnikovskiy:
[Link to the ticket](https://mirantis.jira.com/browse/PRODX-21766)
Now we can Selective Sync mode is working. We can select namespaces and update clouds 
image

### DIFF
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -746,7 +746,7 @@ export class ExtendedCloud extends EventDispatcher {
       this.error = getErrorMessage(this.cloud.connectError);
       logger.error(
         'ExtendedCloud.reconnect()',
-        `Failed to get cloud.connect, error="${this.error}", extCloud=${this}`
+        `Cloud connection failed, error="${this.error}", extCloud=${this}`
       );
     }
   }
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -153,7 +153,7 @@ const _fetchCredentials = async function (cloud, namespaces) {
   //  retrieved for each namespace
   const flattenCreds = flatten(results);
 
-  const error = flattenCreds.find((c) => c.error);
+  const error = (flattenCreds.find((c) => c.error) || {}).error;
   if (error) {
     return { error };
   }
@@ -236,7 +236,7 @@ const _fetchSshKeys = async function (cloud, namespaces) {
     })
   );
 
-  const error = results.find((sk) => sk.error);
+  const error = (results.find((sk) => sk.error) || {}).error;
   if (error) {
     return { error };
   }
@@ -390,7 +390,7 @@ const _fetchClusters = async function (cloud, namespaces) {
     })
   );
 
-  const error = results.find((cl) => cl.error);
+  const error = (results.find((cl) => cl.error) || {}).error;
   if (error) {
     return { error };
   }
@@ -743,7 +743,11 @@ export class ExtendedCloud extends EventDispatcher {
       };
       this.cloud.addEventListener(CLOUD_EVENTS.STATUS_CHANGE, handler);
     } else {
-      this.error = this.cloud.connectError;
+      this.error = getErrorMessage(this.cloud.connectError);
+      logger.error(
+        'ExtendedCloud.reconnect()',
+        `Failed to get cloud.connect, error="${this.error}", extCloud=${this}`
+      );
     }
   }
 
--- a/src/renderer/components/GlobalPage/SyncView.js
+++ b/src/renderer/components/GlobalPage/SyncView.js
@@ -83,9 +83,9 @@ export const SyncView = () => {
    * @param {string} url - cloudUrl
    */
   const getDataToSync = (data, url) => {
-    // store data from each cloud in local object
+    // store data from each cloud in the local object 'syncedClouds'
     setSyncedClouds({ ...syncedClouds, [url]: data });
-    // when clouds count in local object === count ECs preparation is done and we can use syncedClouds
+    // when clouds count in local object === count ECs, preparation is done and we can use syncedClouds
     if (
       Object.keys(syncedClouds).length === Object.keys(extendedClouds).length
     ) {
--- a/src/renderer/components/GlobalPage/SyncView.js
+++ b/src/renderer/components/GlobalPage/SyncView.js
@@ -102,13 +102,13 @@ export const SyncView = () => {
     // this condition happens only when last cloud is written to syncedClouds
     if (!isSyncStarted && Object.keys(syncedClouds).length) {
       // go through all clouds and update properties
-      Object.keys(syncedClouds).map(url => {
-        const {syncAll, syncNamespaces} = syncedClouds[url];
-        const cloud = cloudStore.clouds[url]
+      Object.keys(syncedClouds).map((url) => {
+        const { syncAll, syncNamespaces } = syncedClouds[url];
+        const cloud = cloudStore.clouds[url];
         cloud.syncAll = syncAll;
         cloud.syncNamespaces = syncNamespaces;
-      })
-      closeSelectiveSyncView()
+      });
+      closeSelectiveSyncView();
     }
   }, [syncedClouds, isSyncStarted]);
 
--- a/src/renderer/components/GlobalPage/SyncView.js
+++ b/src/renderer/components/GlobalPage/SyncView.js
@@ -1,4 +1,4 @@
-import { useState, useCallback } from 'react';
+import { useState, useCallback, useEffect } from 'react';
 import { Renderer } from '@k8slens/extensions';
 import styled from '@emotion/styled';
 import { layout } from '../styles';
@@ -61,6 +61,8 @@ export const SyncView = () => {
   } = useExtendedCloudData();
   const [showAddCloudComponent, setShowAddCloudComponent] = useState();
   const [isSelectiveSyncView, setIsSelectiveSyncView] = useState(false);
+  const [isSyncStarted, setIsSyncStarted] = useState(false);
+  const [syncedClouds, setSyncedClouds] = useState({});
 
   const handleAddCloud = useCallback(function (cloud) {
     cloudStore.addCloud(cloud);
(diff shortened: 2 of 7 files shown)


src/renderer/components/hooks/useCheckboxes.js
```diff
@@ -109,8 +109,20 @@ export function useCheckboxes(initialState) {
...
+      return { syncAll: true, syncNamespaces: [] };
+    }
+    return {
+      syncAll: false,
+      syncNamespaces: Object.keys(checkboxesState.children).filter(
+        (name) => checkboxesState.children[name]
+      ),
+    };
+  };
   return {
```
stefcameron, line 123: nit: since this `const getSyncedData()` isn't logically related to the subsequent `return` statement, an empty line here would make the code easier to read.

src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -123,23 +120,15 @@ export const SynchronizeBlock = ({ extendedCloud, onAdd }) => {
...
     const { cloud } = extendedCloud;
-    const allNamespaces = extendedCloud.namespaces.map(({ name }) => name);
-    const namespaces = allNamespaces.filter(
-      (name) => getCheckboxValue({ name }) === checkValues.CHECKED
-    );
+    const { syncNamespaces, syncAll } = getSyncedData();
 
-    if (!namespaces.length) {
+    if (!syncAll && !syncNamespaces.length) {
       Notifications.error(synchronizeBlock.error.noProjects());
```
stefcameron, line 126: If I have 3 clouds being synced, and I go into the Selective Sync mode, and I uncheck 2 of the 3 clouds, will this result in multiple error messages?

Seems like it. Or worse, a subsequent error message will overwrite the first.

I think the notification might work when adding a new Cloud, but when we're in Selective Sync view, that's more like a modal (even though it's not "floating"), so this should generate an error message somewhere in the UI and it should disable the "Synchronize selected projects" button.

Or maybe the Cloud should go red or something like that.

Probably best to work with @mgorczyca-mirantis to figure this out.

And the error message or whatever it is might need to indicate that if they want to remove a Cloud, this isn't the place to do it. They should exit this mode, and then choose the "remove" option from the context menu.

src/renderer/components/GlobalPage/SyncView.js
```diff
@@ -69,6 +71,46 @@ export const SyncView = () => {
...
+    if (!isSyncStarted && Object.keys(syncedClouds).length) {
+      // go through all clouds and update properties
+      Object.keys(syncedClouds).map((url) => {
+        const { syncAll, syncNamespaces } = syncedClouds[url];
+        const cloud = cloudStore.clouds[url];
+        cloud.syncAll = syncAll;
+        cloud.syncNamespaces = syncNamespaces;
+      });
+      closeSelectiveSyncView();
+    }
```
stefcameron, line 112: Will this logic somehow __skip__ updating a Cloud if the user hasn't touched it (hasn't changed its selections) in the Selective Sync mode?

stefcameron (changes requested):
This is simpler than I thought it might be, which is good! That's a testament to well-structured code. 😄 

I think there are a couple of issues, though, around error handling and maybe corner cases, that we need to work out.

src/renderer/components/GlobalPage/SynchronizeBlock.js
bnalisnikovskiy: This is Synchronize (connection) block where we add create Clouds, not a Selective sync

src/renderer/components/GlobalPage/SyncView.js
bnalisnikovskiy: Not, we check all clouds. But I belive our checks in setters, as `newValue !==  _syncAll`, have to handle this

src/renderer/components/hooks/useCheckboxes.js
bnalisnikovskiy: Perhaps we should configure `padding-line-between-statements` eslint rule to cover those cases

src/renderer/components/GlobalPage/SyncView.js
bnalisnikovskiy: Update: `!==` works for bool (expected) but not for arrays, even when nothing was changed inside. So instead that I use lodash `isEqual`. It covers arrays

src/renderer/components/hooks/useCheckboxes.js
stefcameron: Thanks. I checked out that rule, and even though it has a lot of configuration options, in the end, there is no rule that could ever properly decide where an empty line should be added, IMO. I think that's why even Prettier doesn't go there. Unless maybe it was backed with some AI engine of some sort. 😉 

`<SoapBox>`

The way I look at this is that writing code is much like writing a book or blog post or article. Imagine if you wrote a book and didn't create paragraphs. You just had one solid block of words with sentences but no paragraphs, no chapters, just sentences, one after another. That would probably be the most difficult book to read -- and understand.

Paragraphs group sentences into common "chunks" or "ideas", and even though the paragraphs would all be there in that solid block of words, they would be extremely difficult to pick out because you'd have to constantly be _consciously_ thinking about where a paragraph (or specific idea) might have ended, and new one started.

Enter whitespace, and in particular, empty lines (I could go on about whitespace within statements, but I'll save some soap in my soap box for next time...). They are there to create a natural separation between ideas or logical groupings of sentences that are related to one another. And even more so, they provide a _subconscious_ way (i.e. help) for your brain to pick out those chunks of ideas or groupings while you're reading. It makes reading a book a lot more pleasant, and easy.

The same goes for reading code. If we just wrote everything as one big, long stream of statements, it would be super hard to read and understand. But with empty lines, we can create logical groupings of statements to make it easier for someone else to comprehend what's going on, and to skip from one group of statements to another, without focusing on every detail, while still understanding what's going on.

I don't think Prettier or ESLint could come-up with a way to do that which wouldn't look "machined" (or even arbitrary, since the tool doesn't truly understand the code) like that other rule we added last week which just blindly states that all class members shall have an empty line between them. At least in a class declaration, that makes sense most of the time (though I still hold a candle to grouping getters and setters together since that seems logical, but I could see that it might be good to have an empty line between a long getter and the following setter, so I concede to the rule).

<details>
<summary>What the above looks like without paragraphs</summary>
<br/>
The way I look at this is that writing code is much like writing a book. Imagine if you wrote a book and didn't create paragraphs. You just had one solid block of words with sentences but no paragraphs, no chapters, just sentences, one after another. That would probably be the most difficult book to read -- and understand. Paragraphs group sentences into common "chunks" or "ideas", and even though the paragraphs would all be there in that solid block of words, they would be extremely difficult to pick out because you'd have to constantly be _consciously_ thinking about where a paragraph (or specific idea) might have ended, and new one started. Enter whitespace, and in particular, empty lines (I could go on about whitespace within statements, but I'll save some soap in my soap box for next time...). They are there to create a natural separation between ideas or logical groupings of sentences that are related to one another. And even more so, they provide a _subconscious_ way (i.e. help) for your brain to pick out those chunks of ideas or groupings while you're reading. It makes reading a book a lot more pleasant, and easy. The same goes for reading code. If we just wrote everything as one big, long stream of statements, it would be super hard to read and understand. But with empty lines, we can create logical groupings of statements to make it easier for someone else to comprehend what's going on, and to skip from one group of statements to another, without focusing on every detail, while still understanding what's going on. I don't think Prettier or ESLint could come-up with a way to do that which wouldn't look "machined" (or even arbitrary, since the tool doesn't truly understand the code) like that other rule we added last week which just blindly states that all class members shall have an empty line between them. At least in a class declaration, that makes sense most of the time (though I still hold a candle to grouping getters and setters together since that seems logical, but I could see that it might be good to have an empty line between a long getter and the following setter, so I concede to the rule).
</details>

`</SoapBox>`

src/common/ExtendedCloud.js
```diff
@@ -738,7 +738,11 @@ export class ExtendedCloud extends EventDispatcher {
       };
       this.cloud.addEventListener(CLOUD_EVENTS.STATUS_CHANGE, handler);
     } else {
-      this.error = this.cloud.connectError;
+      this.error = getErrorMessage(this.cloud.connectError);
+      logger.error(
+        'ExtendedCloud.reconnect()',
+        `Failed to get cloud.connect, error="${this.error}", extCloud=${this}`
```
stefcameron, line 744:
```suggestion
        `Cloud connection failed, error="${this.error}", extCloud=${this}`
```

src/renderer/components/GlobalPage/SynchronizeBlock.js
bnalisnikovskiy: So we allow user to deselect all projects (`syncAll: false, syncedNamespaces: []`), but display confirmation dialog. Do I understand that correctly?

stefcameron: There are destructive 2 cases we need to deal with, and 3 scenarios:

1. The user deselects everything for a Cloud (`syncAll: false, syncedNamespaces: []`): This is __not allowed__, and basically an error. So the "Synchronize selected projects" button should be __disabled__ -- but then the question arises, how do we inform the user as to why that button is disabled? That's a question for @mgorczyca-mirantis 
2. The user deselects a previously-selected project (so maybe `syncAll` was true and now false and `syncedNamespaces` is no longer an empty array, or at least one item in `syncedNamespaces` has been __removed__): This is allowed, but will result in items disappearing from the catalog, which is something we warn about (and ask confirmation for) when disconnecting entirely from a Cloud, so we should probably warn about it here too -- but how do we do that in a "good" way in that dialog when there could be removals in more than one Cloud? Another question for @mgorczyca-mirantis 
3. If all the user does is __add__ to the existing selection of any Cloud (they might add to the selection of one or more Clouds all at once in this mode), then we don't need to confirm anything because nothing will be removed from the catalog as we will only add to the catalog new items for the new projects we're about to sync.

stefcameron: Thinking about this some more, I think we could apply similar logic to the "Synchronize selected projects" button when adding a new mgmt cluster too: Right now, we leave the button enabled and show an error notification stating at least one project needs to be selected when you deselect all projects and click it. But it seems like the button shouldn't be enabled _unless_ you have at least one project selected... And then @mgorczyca-mirantis is that enough, just disabling the button, or should there be some other type of indication somewhere about the reason why the button is disabled (because there are no projects selected to sync when adding a new mgmt cluster)?

bnalisnikovskiy: Can we merge this PR without those improvements, just with basic logic? 
Fixing each-cases will be in separate ticket

stefcameron: Here's a summary of our UX discussions today regarding this.

__Decisions:__

- Can add a mgmt cluster _without_ selecting any projects. This will result in only mgmt cluster item in Catalog (via SyncManager), whether the checkbox next to it is checked or not.
- When adding a new mgmt cluster, there will be an additional new checkbox to say "auto-sync all future projects" (or similar; mockup will follow)
- Can go into Selective Sync mode and uncheck _everything_ from a mgmt cluster. This will remove all project-related items, but still, the mgmt cluster item will remain in the Catalog (but only that one item since all projects are deselected in this case, and even the checkbox next to the mgmt cluster itself will be unchecked in this state).
- Selective Sync mode will get a new "Auto-sync" column with a checkbox and label for enabling/disabling auto-sync mode, but only for mgmt cluster rows. This checkbox is the direct state of `Cloud.syncAll` boolean.

__TBD:__

- How we summarize what changes will happen when user clicks on "Apply" button in Selective Sync mode is still TBD.

stefcameron: > Can we merge this PR without those improvements, just with basic logic? Fixing each-cases will be in separate ticket

Yes, we can do that. I will create a ticket today for the work with some mocks.

stefcameron (approved):

src/renderer/components/GlobalPage/SynchronizeBlock.js
stefcameron: https://mirantis.jira.com/browse/PRODX-21945

john-papadami: > 

Would also like to add to the auto-sync bullet is that when in selective sync, if one of the mgmnt cluster projects is deselected, auto-sync will also get deslected to indicate that all projects need to be selected in order for auto-sync to work.

Michal and I are meeting today to discuss the TBDs. Once we come to a proposal we will post it in the PR.

PR #734, opened 2022-02-15
Added correct project to confirm message

AntonNos:
Added project list from `extendedCloud.cloud.syncNamespaces` instead of `extendedCloud.namespaces`, because in case we connected not all list of project, we still saw full list of project to related cloud.


bnalisnikovskiy (approved):
LGTM

stefcameron:
Good catch! Though we should probably have used `ExtendedCloud.syncedNamespaces` instead. That property gives you only the synced ones, not all of them.

PR #735, opened 2022-02-15
[PRODX-21561]: added sync now logic

AntonNos:


src/common/ExtendedCloud.js
```diff
@@ -603,6 +603,11 @@ export class ExtendedCloud extends EventDispatcher {
     this.dispatchEvent(EXTENDED_CLOUD_EVENTS.FETCH_DATA, this);
   };
 
+  /** Called when this instance is being updated/synced. */
+  syncData = () => {
```
stefcameron, line 607:
```suggestion
  /** Trigger an immediate data fetch outside the normal fetch interval. */
  fetchNow() {
```

There's no need for a bound property that's a function in this case since it's not an event handler like `onCloudSyncChange()` or `onFetchData()`.

Also, the EC doesn't sync anything; it just fetches data. The SyncManager will sync stuff from the EC to the Lens Catalog.

stefcameron (approved):
This is good! I'd just request a simple change before you merge.

PR #736, opened 2022-02-15
PRODX-21857 Added one more status for EC.fetchingData

bnalisnikovskiy:
Since our Spinner component doesn't suite to the table, I added one more status 'Updating' during fetchingData event. 
In my opinion, spinner or animation would be better, but it's need more work from Lens/designers side. Open to proposals

Also we hide accordion arrow til we get namespaces (in first time)

image

### DIFF
--- a/src/strings.ts
+++ b/src/strings.ts
@@ -332,14 +332,13 @@ export const extendedCloud: Dict = {
 export const connectionStatuses: Dict = {
   cloud: {
     connected: () => 'Connected',
-    connecting: () => 'Connecting',
+    connecting: () => 'Connecting...',
     disconnected: () => 'Disconnected',
-    updating: () => 'Updating',
+    updating: () => 'Updating...',
   },
   namespace: {
     connected: () => 'Synced',
     disconnected: () => 'Not synced',
-    updating: () => 'Updating',
   },
 };
 
--- a/src/strings.ts
+++ b/src/strings.ts
@@ -334,10 +334,12 @@ export const connectionStatuses: Dict = {
     connected: () => 'Connected',
     connecting: () => 'Connecting',
     disconnected: () => 'Disconnected',
+    updating: () => 'Updating',
   },
   namespace: {
     connected: () => 'Synced',
     disconnected: () => 'Not synced',
+    updating: () => 'Updating',
   },
 };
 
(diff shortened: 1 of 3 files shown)


src/strings.ts
```diff
@@ -326,10 +326,12 @@ export const connectionStatuses: Dict = {
     connected: () => 'Connected',
     connecting: () => 'Connecting',
```
stefcameron, line 327: I think it would be better if we had an ellipsis here, "Connecting..."

src/strings.ts
```diff
@@ -326,10 +326,12 @@ export const connectionStatuses: Dict = {
...
     disconnected: () => 'Disconnected',
+    updating: () => 'Updating',
   },
   namespace: {
     connected: () => 'Synced',
     disconnected: () => 'Not synced',
+    updating: () => 'Updating',
```
stefcameron, line 334: I don't think we should go to this level because we don't have per-namespace event notifications and syncing. We just fetch/sync at the cloud level. So for a namespace, the most important status is either "Synced" or "Not synced" based on the Cloud's status ("Disconnected" or not).

stefcameron (changes requested):
Thanks for addressing these two issues!

So with this change, when Lens loads and there's a Cloud in storage, the status column will go from "Disconnected -> Connecting... -> Connected -> Updating... -> Connected", and then 5 minutes later, when we do a scheduled fetch of that Cloud's state, the status column will go from "Connected -> Updating... -> Connected"?

bnalisnikovskiy:
> Thanks for addressing these two issues!
> 
> So with this change, when Lens loads and there's a Cloud in storage, the status column will go from "Disconnected -> Connecting... -> Connected -> Updating... -> Connected", and then 5 minutes later, when we do a scheduled fetch of that Cloud's state, the status column will go from "Connected -> Updating... -> Connected"?

Yes. But after Connecting we see Updating without  'Connected' step. It's related to how and at what moment we get statuses from EC

stefcameron (approved):
Thanks for the updates! 👌

stefcameron:
Note that on Slack, there's a recommendation/expectation/assumption from John that "Disconnected" label is gray, not red. Could you update that before you merge? Use a gray from Lens that's bright enough that provides enough contrast so that the word is easily read.

bnalisnikovskiy:
> Note that on Slack, there's a recommendation/expectation/assumption from John that "Disconnected" label is gray, not red. Could you update that before you merge? Use a gray from Lens that's bright enough that provides enough contrast so that the word is easily read.

Changed to gray. Notice that in figma we have it red. Probably need to be updated. FYI @mgorczyca-mirantis

PR #743, opened 2022-02-17
Fetch all entity types from ExtendedCloud (and more)

stefcameron:
The bulk of this change centers around the following:

- Establish the minimum RTV typesets for our expectations for
  each entity type retrieved from the API.
- Convert `ExtendedCloud.fetchData()` to a "best-try" process
  where we log errors and disregard entities instead of stopping
  the fetch and trying to have an error message. This is in part
  to greatly simplify the fetch, but also to deal with MCC
  disabled components and user permissions that lead to 403
  "access denied" responses when fetching certain entities. If at
  all possible, we don't want to have to replicate this entire
  system from MCC into the extension, so by simply being ready
  to encounter 403s and ignore them (e.g. for BareMetal features
  typically disabled), my hope is to skirt the issue.
- Rename a few key properties that weren't consistently named
  anymore with how the code has evolved.

Also:

- Each API type class (Namespace, Cluster, Proxy, etc.) is now
  more formal with getters instead of plain properties, and
  has a toString() for logging/debugging.
- Each API type class also defines a typeset, like it did before,
  but I reduced the typesets to only the data properties that are
  necessary for the information we need. Best to only care about
  what we need than to check for additional stuff which is
  inconsequential and might change on us, causing issues we
  don't need.
- Added an optimization to ExtendedCloud with a new `preview`
  parameter on the constructor to signal that the instance is
  for preview purposes only, meaning we skip retrieving proxies
  and licenses to save a bit of time. We use this in AddCloudInstance
  where the EC is just throw-away for preview purposes.

### DIFF
--- a/src/common/ExtendedCloud.js
+++ b/src/common/ExtendedCloud.js
@@ -1,15 +1,16 @@
 import * as rtv from 'rtvjs';
 import { Cloud, CLOUD_EVENTS } from './Cloud';
-import { filter, flatten } from 'lodash';
+import { filter } from 'lodash';
 import { cloudRequest, extractJwtPayload } from '../api/apiUtil';
-import * as strings from '../strings';
 import { Namespace } from '../api/types/Namespace';
-import { Credential, credentialTypesList } from '../api/types/Credential';
+import { Credential } from '../api/types/Credential';
 import { SshKey } from '../api/types/SshKey';
-
-import { logger } from '../util/logger';
 import { Cluster } from '../api/types/Cluster';
+import { Proxy } from '../api/types/Proxy';
+import { License } from '../api/types/License';
+import { logger } from '../util/logger';
 import { EventDispatcher } from './EventDispatcher';
+import { apiEntities, apiCredentialEntities } from '../api/apiConstants';
 
 export const EXTENDED_CLOUD_EVENTS = Object.freeze({
   /**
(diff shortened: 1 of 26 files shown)


stefcameron:
Sorry for the mega PR again. Most of it is actually just moving existing code around and renaming a few key properties, aside from the main part which is the massive change to `ExtendedCloud.fetchData()`.

I basically created generic `_fetchEntityList()` and `_deserializeEntityList()` functions that work for all types (Namespace, Cluster, Credential, Proxy, SshKey, License) and then removed a lot of duplicated code in the specialized `_fetchCredentials()`, `_fetchSshKeys()`, `_fetchClusters()`, etc.

With all that done, now the simple types are reduced from 2 functions each with multiple lines of code to just this, for example:

```javascript
const _fetchLicenses = async function (cloud, namespaces) {
  const { entities: licenses, tokensRefreshed } = await _fetchEntityList({
    entity: apiEntities.RHEL_LICENSE,
    cloud,
    namespaces,
    create: ({ data, namespace }) => new License(data, namespace),
  });
  return { licenses, tokensRefreshed };
};
```

Much simpler to read -- and to maintain -- since the bulk of the stuff happens in the 2 generic functions now.

stefcameron:
I still have to do another PR specifically related to all the Catalog entities in `./catalog/...` and add the missing License and Proxy entities. That should be much smaller, so hopefully I can whip that off tomorrow, and you guys will be all set to take on the SyncManager. 😅

src/common/ExtendedCloud.js
```diff
@@ -77,336 +78,344 @@ const getErrorMessage = (error) => {
...
+ * Deserialize a raw list of API entity objects into `ApiObject` instances.
+ * @param {string} entity API entity name to fetch from the `apiConstants.apiEntities` enum.
+ * @param {Object} body Data response from `/list/{entity}` API.
+ * @param {Namespace} [namespace] The related Namespace; `undefined` if fetching
+ *  namespaces themselves (i.e. `entity === apiEntities.NAMESPACE`).
+ * @returns {Array<ApiObject>} Array of API objects. Empty list if none. Any
+ *  item that can't be deserialized is ignored and not returned in the list.
  */
-const _deserializeCredentialsList = function (body, namespace, credentialType) {
+const _deserializeEntityList = function (entity, body, namespace, create) {
```
bnalisnikovskiy, line 89: Nit: `create` is missed in JSdoc

src/common/ExtendedCloud.js
```diff
@@ -651,76 +674,47 @@ export class ExtendedCloud extends EventDispatcher {
...
-
-      if (nsResults?.error) {
-        logger.error(
-          'ExtendedCloud.fetchData()',
-          `Failed to fetch namespaces; error="${nsResults.error}", extCloud=${this}`
-        );
-      }
+    let proxyResults;
+    let licenseResults;
+    if (!this.preview) {
```
bnalisnikovskiy, line 689: Weird, but my IDE can't get path for this.preview. At the same time all other `this._something_` are ok in this method. Probably IDE issue
image

bnalisnikovskiy (approved):
Looks very good!

src/common/ExtendedCloud.js
stefcameron: Strange indeed! The property is definitely there. Gotta get VSCode... 😉

PR #744, opened 2022-02-18
Reconcile custom Catalog entities with API Objects, add License

stefcameron:
This update adds the missing LicenseEntity for adding licenses to the Catalog.

It also establishes all the typesets for expected shapes of catalog entities
of all the custom types.

It removes all the creation-related context menu items since those aren't
part of MVP (except maybe "create cluster", but let's see if we get there).

And finally, it adds a `toEntity()` method to each of the API types
(Cluster, Credential, Proxy, SshKey, License, Namespace) to use when
adding items to the Catalog:

```
catalogSource.push(new ProxyEntity(proxy.toEntity()));
```

### Preview

All entities in the Catalog (dummy data), including the new LicenseEntity, except for clusters since those require kubeconfigs on disk and that's hard to fake.

Screen Shot 2022-02-17 at 7 24 21 PM


src/catalog/LicenseEntity.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (this.metadata.source === consts.catalog.source) {
+      context.menuItems.push({
+        title: `(WIP) ${strings.catalog.entities.common.contextMenu.browserOpen.title()}`,
+        icon: 'launch', // NOTE: must be a string; cannot be a component that renders an icon
+        onClick: async () => {
+          logger.log(
+            'LicenseEntity/LicenseEntity.onContextMenuOpen.browserOpen',
+            'opening License %s in browser...',
+            this.metadata.name
+          );
```
stefcameron, line 92: @AntonNos These context menu items (for each item type) also need to be implemented as part of steps 13 and 14 in the roadmap.

bnalisnikovskiy (approved):
Wow, looks very good. 

As I can see you removed almost all menu items from namespace (project). Do we plan add some of them back in the future?

bnalisnikovskiy:
Also, I noticed that we have to remove `cluster-store.json` because the structure inside doesn't match the new one. Is it okay?

PR #748, opened 2022-02-18
PRODX-21945 Fixed Selective sync UX issues

bnalisnikovskiy:
[Link to the ticket](https://mirantis.jira.com/browse/PRODX-21945)
Main changes: 
- Added 'Synchronize future projects' checkboxes for `syncAll` bool in SynchronizeBlock and in SelectiveSync Table
- Added `Cloud.ignoredNamespaces` where we store all unchecked projects
- Added logic to ExtendedCloud, which updates `Cloud.ignoredNamespaces`  and `Cloud.syncedNamespaces` in it got new project. Depending on `syncAll` we're choosing where to move new project ignored or synced
- Fixed issue when SelectiveSync table isn't updated after new data come (eg when Cloud disconnected=> we reconnect it and go to SelectiveSync view before EC.dataFetch is finished)

image
image

### DIFF
--- a/src/common/Cloud.js
+++ b/src/common/Cloud.js
@@ -476,7 +476,7 @@ export class Cloud extends EventDispatcher {
       },
     });
 
-    Object.defineProperty(this, 'updateNamespaces', {
+    Object.defineProperty(Object.getPrototypeOf(this), 'updateNamespaces', {
       enumerable: true,
       /**
        * @param {Array<string>} syncedList A list of namespace names in the mgmt cluster that should be synced.
@@ -538,6 +538,13 @@ export class Cloud extends EventDispatcher {
     }
 
     return CONNECTION_STATUSES.DISCONNECTED;
+  }
+
+  /**
+   * @member {Array<string>} returns syncedNamespaces + ignoredNamespaces array
+   */
+  get allNamespaces() {
+    return this.syncedNamespaces.concat(this.ignoredNamespaces)
   }
 
   /**
--- a/src/common/Cloud.js
+++ b/src/common/Cloud.js
@@ -172,6 +172,7 @@ export class Cloud extends EventDispatcher {
     cloudUrl: [rtv.EXPECTED, rtv.STRING],
 
     syncedNamespaces: [rtv.EXPECTED, rtv.ARRAY, { $: rtv.STRING }],
+    ignoredNamespaces: [rtv.EXPECTED, rtv.ARRAY, { $: rtv.STRING }],
     name: [rtv.EXPECTED, rtv.STRING],
     syncAll: [rtv.EXPECTED, rtv.BOOLEAN],
   };
@@ -190,6 +191,7 @@ export class Cloud extends EventDispatcher {
     let _name = null;
     let _syncAll = false;
     let _syncedNamespaces = [];
+    let _ignoredNamespaces = [];
     let _token = null;
     let _expiresIn = null;
     let _tokenValidTill = null;
@@ -242,8 +244,7 @@ export class Cloud extends EventDispatcher {
 
     /**
      * @member {Array<string>} syncedNamespaces A list of namespace names in the mgmt
-     *  cluster that should be synced. If syncAll is true, this list is ignored and
-     *  all existing/future namespaces are synced.
+     *  cluster that should be synced.
      */
     Object.defineProperty(this, 'syncedNamespaces', {
       enumerable: true,
@@ -258,6 +259,29 @@ export class Cloud extends EventDispatcher {
           );
         if (!isEqual(_syncedNamespaces, newValue)) {
           _syncedNamespaces = newValue;
+          this.dispatchEvent(CLOUD_EVENTS.SYNC_CHANGE, this);
+        }
+      },
+    });
+
+    /**
+     * @member {Array<string>} ignoredNamespaces A list of namespace names in the mgmt
+     *  cluster that not synced.
+     */
+    Object.defineProperty(this, 'ignoredNamespaces', {
+      enumerable: true,
+      get() {
+        return _ignoredNamespaces;
+      },
+      set(newValue) {
+        DEV_ENV &&
+          rtv.verify(
+            { token: newValue },
+            { token: Cloud.specTs.ignoredNamespaces }
+          );
+        if (!isEqual(_ignoredNamespaces, newValue)) {
+          _ignoredNamespaces = newValue;
+          // not sure about event type for this one. Probably we need to add a new one
           this.dispatchEvent(CLOUD_EVENTS.SYNC_CHANGE, this);
         }
       },
@@ -598,6 +622,7 @@ export class Cloud extends EventDispatcher {
       this.name = spec.name;
       this.syncAll = spec.syncAll;
       this.syncedNamespaces = spec.syncedNamespaces;
+      this.ignoredNamespaces = spec.ignoredNamespaces;
       this.username = spec.username;
 
       if (spec.id_token && spec.refresh_token) {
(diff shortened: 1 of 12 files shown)


stefcameron:
👀

src/renderer/components/hooks/useCheckboxes.js
```diff
@@ -15,26 +15,33 @@ const setParentCheckboxState = (children) => {
...
+  }
+  // otherwise, we make state for all EC.namespaces, they have to be present
+  // And use syncedNamespaces to make initial state checked for selected checkboxes
+  return extCloud.namespaces.reduce((acc, namespace) => {
+    acc[namespace.name] = extCloud.cloud.syncedNamespaces.includes(
+      namespace.name
+    );
     return acc;
   }, {});
 };
```
stefcameron, line 37: Somewhere, not sure if it's here specifically, there's a few bugs when the Cloud is __disconnected__ and I'm going in/out of Selective Sync mode.

- First, when I'm in SyncView (not Selective Sync), I don't see the twisty with the list of synced namespaces (just without the cluster/cred/key counts):

image

- Next, if I go into Selective Sync mode and simply toggle the `syncAll` flag from true to false (I uncheck it), I lose the entire list of `ignoredNamespaces` on the Cloud (and in the `cloud-store.json`).

- If I go into Selective Sync mode and uncheck a project and apply changes, it correctly moves into `ignoredNamespaces`. But if I also toggle the `syncAll` flag true or false, I will lose that whole list.

- If I uncheck all namespaces, so they all move to `ignoredNamespaces`, and then go back into Selective Sync view, I don't see any namespaces. When the Cloud is disconnected, we should show the full list of namespaces from `syncedNamespaces` and `ignoredNamespaces`, with them checked/unchecked, respectively.
    - Even though we're disconnected, we shouldn't forget about ignored namespaces -- especially if the user has `syncAll: true`. Otherwise, once we finally reconnect, we'll end-up auto-syncing all of the ones they deselected just minutes ago.
    - When we're disconnected, removing a namespace will cause items to disappear from the Catalog, which is fine.
    - When we're disconnected, adding a namespace will simply have no effect until the Cloud is reconnected again. The SyncManager will just do nothing (since `EC.syncedNamespaces` won't reflect that namespace until the Cloud is reconnected anyway).

- We're missing the "(Disconnected)" label next to the Cloud's name in Selective Sync view.

- If a Cloud doesn't have any namespaces, then the twisty shouldn't be visible in Selective Sync view like it is today:

image

src/strings.ts
```diff
@@ -85,13 +86,15 @@ export const syncView: Dict = {
...
   connectButtonLabel: () => 'Connect new Management Cluster',
+  autoSync: () => 'auto-sync',
 };
 export const managementClusters = {
   table: {
     thead: {
       name: () => 'Name',
       url: () => 'URL',
       username: () => 'Username',
+      autosync: () => 'Autosync',
```
stefcameron, line 97: I know the mock-up shows "Autosync", but it should be "Auto-sync".

src/strings.ts
```diff
@@ -85,13 +86,15 @@ export const syncView: Dict = {
   synchronizeProjectsButtonLabel: () => 'Synchronize selected projects',
   syncButtonLabel: () => 'Selective sync...',
   connectButtonLabel: () => 'Connect new Management Cluster',
+  autoSync: () => 'auto-sync',
```
stefcameron, line 89: Let's capitalize it, "Auto-sync"

stefcameron (changes requested):
Nice work! You hit every nail on the head in the ticket except for a missing label. 🎉 I did some local testing and found a few issues when the Cloud is disconnected. See comments.

src/common/Cloud.js
```diff
@@ -475,6 +476,37 @@ export class Cloud extends EventDispatcher {
       },
     });
 
+    Object.defineProperty(this, 'updateNamespaces', {
```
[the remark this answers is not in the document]
bnalisnikovskiy: Done

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -404,6 +405,8 @@ EnhancedTableRow.propTypes = {
   withCheckboxes: PropTypes.bool,
   isSyncStarted: PropTypes.bool,
   getDataToSync: PropTypes.func,
+  namespaces: PropTypes.arrayOf(PropTypes.object).isRequired,
```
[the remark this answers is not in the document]
bnalisnikovskiy: 'namespaces' is always an object. It could be a class or just `{name: 'foo' }`. We make it in `TableRowListenerWrapper` => `getNamespaces`.
  Updated

src/renderer/components/hooks/useCheckboxes.js
bnalisnikovskiy: Improved. I hope I cover all cases. 
Strange but I couldn't reproduce your issues with `syncAll` and empty `ignoredNamespaces`

stefcameron: You got them all except the last one: In Selective Sync mode, when disconnected (though this may not make a difference; should test both connected and disconnected) and the Cloud does not have any namespaces (synced and ignored lists are both empty), the twisty is still visible (but you can't click it):

image

The twisty should be hidden in this case, because there's nothing to reveal related to the Cloud.

stefcameron:
One last thing (aside from the twisty issue I pointed out): We should always sort namespaces alphabetically based on name, with the exception of always putting the "default" namespace at the top of the list.

So if we had default, a-namespace, and some-namespace, they would be listed as:

- default
- a-namespace
- some-namespace

The default namespace is where MCC puts the mgmt cluster, so it's treated a little bit special. It always exists since it's the default namespace in Kubernetes.

Let's do this only when we present the namespaces in the UI (in components), not in Cloud or ExtendedCloud since they don't need them to be sorted.

bnalisnikovskiy:
Fixed last issue and added sorting for namespaces

src/renderer/components/EnhancedTable/tableUtil.js
```diff
@@ -44,5 +44,29 @@ export const sortData = (obj, sortBy, order, path) => {
...
+  }
+  if (nameA < nameB) {
+    return -1;
+  }
+  if (nameA > nameB) {
+    return 1;
+  }
+
+  return 0;
+};
```
stefcameron, line 69: `String.includes()` does not do an exact match. We should look for the "default" namespace _exactly_, otherwise, the sort will get messed-up if a namespace includes the word "default". There can only be one "default" namespace.

Also, using `String.localeCompare()` might be safer than a straight numerical comparison.

So this could be as simple as:

```suggestion
const compareNamespaces = (first, second) => {
  if (first.name === 'default') {
    return -1;
  }

  if (second.name === 'default') {
    return 1;
  }

  return first.name.localeCompare(second.name);
};
```

bnalisnikovskiy: I wasn't sure about 'default' thanks. Good point, about `localeCompare` appreciated

stefcameron (approved):
I think we've got it! Thanks for fixing all those bugs.

There's just 2 more little things, but no need to re-review, just merge after fixing. In Selective Sync view:

Screen Shot 2022-02-22 at 12 19 45 PM

- [ ] When the Cloud is disconnected, the "(Disconnected)" label next to the name doesn't look right. Should be `asdf (Disconnected)` in the image above.
- [ ] If the Cloud doesn't have any namespaces to show, we've removed the twisty (which is good), but we should also make sure the checkbox (which would otherwise "select all/none" namespaces) is both disabled and unchecked. So it's still visible, just unchecked and you can't check it.

Once you merge this, we can declare steps 10 and 11 "done" and I'll post another alpha build to NPM for people to play with.

PR #753, opened 2022-02-21
[PRODX-21978]: Changed namespace status to 'Synced' while Cloud is updating

AntonNos:
Screenshot 2022-02-21 at 12 26 04


bnalisnikovskiy (approved):
Looks good and simple. It should be enough for our case :+1:

stefcameron (approved):
👍

PR #756, opened 2022-02-22
WIP [PRODX-22032]: added entities details

AntonNos:


src/store/ClusterStore.js
```diff
@@ -0,0 +1,114 @@
+//
```
bnalisnikovskiy, line 1: this file should be removed

src/renderer/components/EntitiesDetails/MccCredentialEntity.js
```diff
@@ -0,0 +1,35 @@
...
+    </>
+  );
+};
+
+MccCredentialEntity.propTypes = {
+  data: PropTypes.object.isRequired,
+  provider: PropTypes.string,
+  region: PropTypes.string,
+  status: PropTypes.string,
+  created: PropTypes.string,
```
bnalisnikovskiy, line 27: It doesn't work in that way. Check propTypes docs. You need `shape` here

bnalisnikovskiy (changes requested):

stefcameron (changes requested):
@AntonNos I appreciate your work here, but this seems much too complicated for what actually needs to be done.

Just move [this array](https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer/renderer.tsx#L105-L198) into a new module, say, `/src/renderer/catalogEntityDetails.js`

Export the array from the module and import it into renderer.tsx and assign it at line 105.

Then all that remains to do is just add the missing properties, which means adding a few more `<DrawerItem>`s, a few more strings, and related property typesets to the relevant Catalog entity class. For example, if you're adding an additional property for a CredentialEntity, you would add that property to the `spec` object defined here: https://github.com/Mirantis/lens-extension-cc/blob/master/src/catalog/CredentialEntity.ts#L48

No need to move any mock data out of `IpcMain`. SyncManager work will take care of that. @AntonNos @bnalisnikovskiy see my suggestion for how to approach the SyncManager work here: https://mirantis.jira.com/browse/PRODX-21909?focusedCommentId=886612

That's pretty much it. It should be a relatively simple change.

Also, I'm concerned that you're bringing back old things like the `cluster-store` that was deleted. Please make sure you're synced to the latest master.

stefcameron:
Closing in favor of #763

PR #758, opened 2022-02-23
PRODX-22081 Updated extractJwtPayload and add missed Credential type

bnalisnikovskiy:
[Link to the ticket](https://mirantis.jira.com/browse/PRODX-22081)
- Used Buffer instead of `atob` for Node.js env
- Added one more Credential kind to the validation list


stefcameron (approved):
👌

PR #759, opened 2022-02-23
PRODX-22043 [WIP] Add SyncManager

bnalisnikovskiy:


src/store/SyncStore.js
```diff
@@ -0,0 +1,151 @@
...
+    rtv.EXPECTED,
+    rtv.ARRAY,
+    { $: [rtv.OBJECT, { ctor: licenseEntityModelTs }] },
+  ],
+  proxies: [
+    rtv.EXPECTED,
+    rtv.ARRAY,
+    { $: [rtv.OBJECT, { ctor: proxyEntityModelTs }] },
+  ],
+};
```
stefcameron, line 34: This is a good try, but the `OBJECT` type doesn't take a `ctor` ("constructor") option because the `OBJECT` type doesn't validate a class instance. Only `CLASS_OBJECT` does. The docs for the [shape args](https://rtvjs.stefcameron.com/api/#rtvref.types.shape_object_args) explains this (though the generated docs, I admit, are not the easiest to read; detailed, I hope, but not easy to read).

So `ctor` is just being ignored here, and the only validation that's happening is that each property is an `EXPECTED` (i.e. could be `null` also, which I don't think we want to have to handle) array of [OBJECT](https://rtvjs.stefcameron.com/api/#rtvref.types.OBJECT) types (which is pretty much anything...).

Since `getDefaults()` initializes the store's properties to empty array, we don't need to handle a `null` case for any of them. That means each property can be `REQUIRED`, and since `REQUIRED` is the default "qualifier", it doesn't need to be mentioned.

As for the arrays, each element should be an object with the specified _shape_ of `credentialEntityModelTs` ("Ts" is short for "Typeset", which is either a shape that describes an object and implies the `OBJECT` type, unless a specific object type is specified, or a list of possible types, which is where the array comes in; a bit difficult to explain, but I hope the [tutorial](https://rtvjs.stefcameron.com/#tutorials) is nice walk through and makes it easy to understand what all the parts are).

So in the end, with `ARRAY` short notation, the entire `storeTs` can be reduced to this:

```javascript
credentials: [[credentialEntityModelTs]],
sshKeys: [[sshKeyEntityModelTs]],
clusters: [[rtv.ANY]], // eventually: [[clusterEntityModelTs]],
licenses: [[licenseEntityModelTs]],
proxies: [[proxyEntityModelTs]],
```

By default, a `REQUIRED ARRAY` does not have a minimum length. It just requires the overall type to be an `Array`, so it can be empty. You could use `ARRAY` type options to specify a minimum length (if you needed to).

src/store/SyncStore.js
```diff
@@ -0,0 +1,151 @@
...
+
+  constructor() {
+    super({
+      configName: 'sync-store',
+      defaults: SyncStore.getDefaults(),
+    });
+    makeObservable(this);
+  }
+
+  /** Reset clouds to default empty object */
```
stefcameron, line 72: "clouds" -> "properties"

stefcameron (commented):
I took a quick peak... 🙂 You're on the right track!

src/store/SyncStore.js
bnalisnikovskiy: I'm wondering why, but signature like this doesn't work. Left a comment on line 11

stefcameron: I missed that. It should work, though. I'd have to see the error you're getting from RTV.js when it fails to verify the shape.

PR #763, opened 2022-02-28
[PRODX-22001] Added details to all catalog items

AntonNos:


src/renderer/catalogEntityDetails.tsx
```diff
@@ -0,0 +1,158 @@
...
+          />
+          <DrawerItem
+            name={strings.catalog.entities.common.details.props.uid()}
+          >
+            {props.entity.metadata.uid || unknownValue()}
+          </DrawerItem>
+          <DrawerItem
+            name={strings.catalog.entities.sshKey.details.props.dateCreated()}
+          >
+            {props.entity.spec.created || unknownValue()}
```
stefcameron, line 46: All of the `props.entity.spec.created` would change to `props.entity.spec.createdAt` and we would feed that into [dayjs.from(createdAt, true)](https://day.js.org/docs/en/display/from), which should give us something like "4 months" or "2 days", etc.

The final value would be `` createdAt ? `${dayjs(createdAt).format('MMMM DD, YYYY')} (dayjs.from(createdAt, true))` : unknownValue() ``

And since we'll repeat this in a bunch of places, might as well make a function out it. 🙂 

__Note:__ We don't have [dayjs](https://day.js.org/en/) as a dependency yet. `yarn add --dev dayjs` and then in this module:

```javascript
import dayjs from 'dayjs';
import dayjsRelativeTimePlugin from 'dayjs/plugin/relativeTime';
dayjs.extend(dayjsRelativeTimePlugin);

const getCreatedValue = function (createdAt) {
  // you can use dayjs.from(createdAt, true)...
}
```

src/renderer/catalogEntityDetails.tsx
```diff
@@ -0,0 +1,158 @@
...
+          </DrawerItem>
+          <DrawerItem
+            name={strings.catalog.entities.credential.details.props.region()}
+          >
+            {props.entity.spec.region || unknownValue()}
+          </DrawerItem>
+          <DrawerItem
+            name={strings.catalog.entities.credential.details.props.mccStatus()}
+          >
+            {props.entity.spec.status || unknownValue()}
```
stefcameron, line 84: Check `props.entity.spec.valid` and use "Valid" if true or "Invalid" if false.

src/renderer/catalogEntityDetails.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+const {
+  catalog: {
+    entities: {
+      common: { details: unknownValue },
+    },
+  },
+} = strings;
+
+const catalogEntityDetails = [
```
stefcameron, line 28: I'd prefer to avoid default exports, so just `export const catalogEntityDetails = ...` here. We don't use default exports in any other module in the project, so let's be consistent.

stefcameron (changes requested):
Thank you! This is going much better than #756 A few comments, but nothing huge. Not much left to do.

src/catalog/CredentialEntity.ts
```diff
@@ -54,8 +54,8 @@ export const credentialEntityModelTs = mergeRtvShapes(
         },
       ],
       region: [rtv.OPTIONAL, rtv.STRING],
-      status: [rtv.OPTIONAL, rtv.STRING],
-      created: [rtv.OPTIONAL, rtv.STRING],
+      valid: rtv.BOOLEAN,
+      createdAt: [rtv.OPTIONAL, rtv.STRING],
```
stefcameron, line 58: `createdAt` shouldn't be optional, so just `createdAt: rtv.STRING`. Also, I had mentioned this in [this comment](https://github.com/Mirantis/lens-extension-cc/pull/763#discussion_r816056652):

> Since every Catalog Entity will have a "Created" line item in the Details panel, the `createdAt` property should be defined once for all entities here in the currently empty `spec: {}` block: https://github.com/Mirantis/lens-extension-cc/blob/master/src/catalog/catalogEntities.js#L51

But I think it may have been hidden by GitHub and you just missed it. I don't like the way GH hides some comments when there are many.

src/catalog/CredentialEntity.ts
```diff
@@ -66,8 +66,8 @@ export const credentialEntityModelTs = mergeRtvShapes(
 export interface CredentialSpec {
   provider?: string;
   region?: string;
-  status?: string;
-  created?: string;
+  valid: boolean;
+  createdAt?: string;
```
stefcameron, line 70: `createdAt` shouldn't be optional, so just `createdAt: string` for the TypeScript typings -- this goes for all other entities as well since they all have `createdAt`.

src/renderer/catalogEntityDetails.tsx
```diff
@@ -41,9 +51,9 @@ const catalogEntityDetails = [
             {props.entity.metadata.uid || unknownValue()}
           </DrawerItem>
           <DrawerItem
-            name={strings.catalog.entities.sshKey.details.props.dateCreated()}
+            name={strings.catalog.entities.common.details.props.dateCreated()}
           >
-            {props.entity.spec.created || unknownValue()}
+            {getCreatedValue(props.entity.spec.createdAt) || unknownValue()}
```
stefcameron, line 56: Since `createdAt` is a required property, this can simply be:

```suggestion
            {getCreatedValue(props.entity.spec.createdAt)}
```

Especially since `getCreatedValue()` will always return a string anyway, so the `|| unknownValue()` will never get called.

Please update this for all occurrences.

src/strings.ts
```diff
@@ -287,7 +287,7 @@ export const catalog: Dict = {
       // details panel
       details: {
         props: {
-          dateCreated: () => 'Date created',
+          //
```
stefcameron, line 290: This empty comment should be removed.

stefcameron (changes requested):
Thanks for the updates! Just a couple of things missed from the previous review.

src/main/IpcMain.js
```diff
@@ -29,6 +29,10 @@ export class IpcMain extends Main.Ipc {
   //
   // SINGLETON
   //
+  constructor(extension) {
+    super(extension);
+    extension.addCatalogSource(consts.catalog.source, catalogSource);
+  }
```
stefcameron, line 35: I ended-up fixing this on `master`; you'll see when you rebase.

stefcameron (approved):
Awesome, thank you! I will merge this.

PR #768, opened 2022-03-03
[PRODX-22247]: added open in browser functionality to catalog entities

AntonNos:


src/util/helper.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Generate URL for single entity.
+ * @param {object} metadata Single entity metadata.
+ */
+export const generateEntityUrl = (metadata) => {
+  return `${metadata.cloudUrl}/${metadata.namespace}/type?name=${metadata.name}`;
```
stefcameron, line 6: Not quite. The URL isn't valid. Use your access to the internal MCC instance to figure out what the URL will be given the entity type.

Also, instead of a new generic "helper" module, let's put this method in `src/catalog/catalogEntities.js`

src/catalog/CredentialEntity.ts
```diff
@@ -111,9 +113,11 @@ export class CredentialEntity extends Common.Catalog.CatalogEntity<
...
       context.menuItems.push({
-        title: `(WIP) ${strings.catalog.entities.common.contextMenu.browserOpen.title()}`,
+        title: strings.catalog.entities.common.contextMenu.browserOpen.title(),
         icon: 'launch', // NOTE: must be a string; cannot be a component that renders an icon
         onClick: async () => {
+          openBrowser(generateEntityUrl(this.metadata));
+
           logger.log(
             'CredentialEntity/CredentialEntity.onContextMenuOpen.browserOpen',
             'opening Credential %s in browser...',
```
stefcameron, line 123: These placeholder log statements can be removed for each entity type.

stefcameron (changes requested):

src/util/helper.js
stefcameron: Also, maybe you can use the enumeration values from https://github.com/Mirantis/lens-extension-cc/blob/master/src/api/apiConstants.js#L11 to help identify the entity for which you need a URL instead of creating another "entity types" list. So we'd have `generateEntityUrl(apiEntities.PUBLIC_KEY, entity)`.

Might also be good for this function to receive the full entity, not just its metadata.

stefcameron: I actually ended-up having to implement something that would help here, but it's still on #759 right now, which is that every Catalog entity will now have a `entity.metadata.kind` property, which will be one of the values in the new `apiKinds` enum, which is now defined in `/api/apiConstants.js` (see the file in the PR).

So with that, you wouldn't need to pass the kind in, because you'd already have it. We would just need to pass the Catalog entity into the function, like `generateEntityUrl(entity)`, and internally, you could compare `entity.metadata.kind` to one of the values in `apiKinds` and generate the correct URL based on that kind.

There is also `apiCredentialKinds` in `/api/apiConstants.js` which provides an enum of just the credentials (i.e. a subset of `apiKinds`), which could help to identify if the entity is some type of credential (the URL will be the same for any type of credential, except its name, of course).

stefcameron: This is now all on master: https://github.com/Mirantis/lens-extension-cc/blob/master/src/api/apiConstants.js#L56

AntonNos: > Also, maybe you can use the enumeration values from https://github.com/Mirantis/lens-extension-cc/blob/master/src/api/apiConstants.js#L11 to help identify the entity for which you need a URL instead of creating another "entity types" list. So we'd have `generateEntityUrl(apiEntities.PUBLIC_KEY, entity)`.
> 
> Might also be good for this function to receive the full entity, not just its metadata.

Thanks Stefan, I fixed your comments. But have only 1 question - there is no 'general' enum for Credentials in `apiEntities`. I used `OPENSTACK_CREDENTIAL` for credential entity. Is it OK or should I add one more enum `CREDENTIAL`?

stefcameron: With what I pushed to master last night, you don't need to pass the entity into the function because you have it in `entity.metadata.kind`, and for credentials, you can check if the `kind` is included in `apiCredentialKinds` values instead of checking for one specific value like the other entities.

stefcameron (changes requested):

src/catalog/catalogEntities.js
```diff
@@ -137,3 +138,30 @@ export const clusterEntityModelTs = mergeRtvShapes({}, catalogEntityModelTs, {
...
+ * @param {object} entity Single entity.
+ */
+export const generateEntityUrl = (entity) => {
+  const url = `${entity.metadata.cloudUrl}/projects/${entity.metadata.namespace}`;
+
+  if (Object.values(apiCredentialKinds).includes(entity.metadata.kind)) {
+    return `${url}/creds?name=${entity.metadata.name}`;
+  }
+
+  switch (entity.metadata.kind) {
```
stefcameron, line 153: Somehow we lost the cluster case. 🙂 We need `case apiKinds.CLUSTER` and it should open to the `/clusters/{entity.metadata.name}` page of the MCC instance.

stefcameron (changes requested):
Aaaaaalllllmoooost there... 😉

src/catalog/catalogEntities.js
AntonNos: Ohh, I didn't add it because had a plan to add it in scope of "Cluster" task. BTW, I added it here:)

stefcameron: Oh, OK, I didn't realize that. But thanks for adding it anyway!

stefcameron (approved):
🎉

PR #797, opened 2022-03-24
[PRODX-22031]: "Clusters" additional information

AntonNos:
- Added additional informations
- Added "Open in browser" option

Please check attached:
First screenshot - non Management cluster panel
Second screenshot - Management cluster panel

Screenshot 2022-03-24 at 10 53 47
Screenshot 2022-03-24 at 10 54 02

### DIFF
--- a/src/renderer/catalogEntityDetails.tsx
+++ b/src/renderer/catalogEntityDetails.tsx
@@ -61,6 +61,57 @@ export const catalogEntityDetails = [
           >
             {getCreatedValue(props.entity.spec.createdAt)}
           </DrawerItem>
+          <DrawerItem
+            name={strings.catalog.entities.cluster.details.props.managementCluster()}
+          >
+            {strings.catalog.entities.cluster.details.props.isManagementCluster(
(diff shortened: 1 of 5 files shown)


src/renderer/catalogEntityDetails.tsx
```diff
@@ -52,15 +52,66 @@ export const catalogEntityDetails = [
...
           <DrawerItem
             name={strings.catalog.entities.common.details.props.dateCreated()}
           >
             {getCreatedValue(props.entity.spec.createdAt)}
           </DrawerItem>
+          <DrawerItem
+            name={strings.catalog.entities.common.details.props.uid()}
+          >
+            {props.entity.metadata.uid || unknownValue()}
+          </DrawerItem>
```
stefcameron, line 114: To be consistent with all the other entities when showing details, let's move UID to the top (first), then created date, and then the rest as you have them.

stefcameron (changes requested):
Nice work! Just a few small details to fix.

PR #806, opened 2022-03-29
[PRODX-22754]: fixed bug with table sorting

AntonNos:
- Fixed issue, when sorting was working only with `URL` column
- Fixed visual behaviour of sorting arrow

### DIFF
--- a/src/renderer/components/EnhancedTable/EnhancedTableHead.js
+++ b/src/renderer/components/EnhancedTable/EnhancedTableHead.js
@@ -25,6 +25,11 @@ const EnhSortButton = styled.button`
   color: var(--textColorAccent);
   background: transparent;
   cursor: pointer;
+
+  i {
+    ${({ isCurrentAsc }) => isCurrentAsc && 'transform: rotate(180deg);'}
+    opacity: ${({ isSortedByCurrent }) => (isSortedByCurrent ? '1;' : '0;')}
+  }
 `;
 
 const sortButtonStyles = {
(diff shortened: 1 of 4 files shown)


src/renderer/components/EnhancedTable/EnhancedTableHead.js
```diff
@@ -25,6 +25,11 @@ const EnhSortButton = styled.button`
   color: var(--textColorAccent);
   background: transparent;
   cursor: pointer;
+
+  i {
+    ${({ isCurrentAsc }) => isCurrentAsc && 'transform: rotate(180deg);'}
+    opacity: ${({ isSortedByCurrent }) => (isSortedByCurrent ? '1;' : '0;')}
```
stefcameron, line 31: I guess this "hides" it or "shows" it. Could we use the "visibility" style here instead? Set `visibility: visible` or `visibility: hidden` as appropriate. If hidden, it'll hide it without removing it from layout, so it'll maintain its presence and things won't get misaligned. Using "opacity" for this feels like a bit of a hack.

src/renderer/components/EnhancedTable/EnhancedTableHead.js
```diff
@@ -65,5 +79,7 @@ export const EnhancedTableHead = ({ sortBy, headerValues }) => {
 
 EnhancedTableHead.propTypes = {
   sortBy: PropTypes.func.isRequired,
+  sortedBy: PropTypes.string.isRequired,
+  order: PropTypes.string.isRequired,
```
stefcameron, line 83: This would be more useful if it was typed as `order: PropTypes.oneOf(['asc', 'desc']).isRequired`

stefcameron (changes requested):

stefcameron:
I don't think I was able to reproduce the bug you were talking about in Slack. AFAICT, all the original issues I found are fixed.

But there is one strange thing, which is that even though the status of both Clouds is "Disconnected", toggling the sort order still changes the order. I would've expected the order to remain stable since the values are the same. But I think this is likely related to `Array.sort()` stability not being guaranteed earlier than ECMAScript 2019, and I'm assuming the version of Electron Lens is using might be earlier than that. I can't see what piece of our code would make this unstable.

Sorting the URL when we have 2 Clouds, one with an IP address and one with a domain, may result in the opposite order you expect, but I think that's because "1" comes before "a" in ASCII.

src/renderer/components/EnhancedTable/EnhancedTableHead.js
AntonNos: Fixed

stefcameron (approved):

PR #815, opened 2022-04-06
[PRODX-23054]: Removed child count from management cluster details view

AntonNos:
Screenshot 2022-04-06 at 13 38 01


stefcameron (approved):
👍

PR #816, opened 2022-04-06
[PRODX-23053]: added proxies and licenses counts for synced namespaced

AntonNos:
Screenshot 2022-04-06 at 14 03 21
Screenshot 2022-04-06 at 14 03 35

@stefcameron Should we show row for entity count if `count === 0` ?

### DIFF
--- a/src/strings.ts
+++ b/src/strings.ts
@@ -102,6 +102,8 @@ export const managementClusters = {
       clusters: () => 'clusters',
       sshKeys: () => 'SSH keys',
       credentials: () => 'credentials',
+      proxies: () => 'proxies',
+      licenses: () => 'licenses',
     },
   },
 };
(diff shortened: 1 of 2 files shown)


stefcameron:
> @stefcameron Should we show row for entity count if count === 0 ?

Yes. My thinking is that by showing the full list with counts, even at 0, but only for synced projects, it's now easy for the customer to get an overview of how many resources of each type are synced into the Lens Catalog for a given project.

src/strings.ts
```diff
@@ -229,7 +231,6 @@ export const catalog: Dict = {
           isManagementCluster: (isManagementCluster) =>
             isManagementCluster ? 'Yes' : 'No',
           url: () => 'URL',
-          childClusters: () => 'Child clusters',
```
stefcameron, line 232: Glad you caught this one too. 😉

stefcameron (approved):
👍

PR #823, opened 2022-04-08
[PRODX-23055]: Dropped showing counts on un-synced/ignored namespaces in all views

AntonNos:
**Synchronize block:**
Screenshot 2022-04-08 at 12 51 33
Screenshot 2022-04-08 at 12 51 48

**Selective sync view**
Screenshot 2022-04-08 at 12 52 07

**Add Cloud Instance view**
Screenshot 2022-04-08 at 13 08 31


stefcameron (approved):
Looks good! I tested locally, seems to work the way it's intended. Thank you! 😄

PR #834, opened 2022-04-18
[PRODX-23052]: added warning when more than 10 projects are selected

AntonNos:
Screenshot 2022-04-18 at 14 32 34
Screenshot 2022-04-18 at 14 32 44
Screenshot 2022-04-18 at 14 33 57


src/renderer/components/WarningIcon.js
```diff
@@ -0,0 +1,27 @@
+import propTypes from 'prop-types';
+
+export const WarningIcon = function ({ size, fill, ...svgProps }) {
```
stefcameron, line 3: Instead of this custom SVG icon, we should be using the Material [warning_amber](https://fonts.google.com/icons?selected=Material%20Icons%3Awarning_amber%3A) icon, which is the closest match. Always use icons from Material, not our own, so that icons match the Lens style.

src/strings.ts
```diff
@@ -78,6 +78,8 @@ export const synchronizeBlock = {
   error: {
     noProjects: () => 'Select at least one project to sync',
   },
+  warning: () =>
+    'Synchronizing large number of projects can cause performance issues',
```
stefcameron, line 82:
```suggestion
    'Synchronizing a large number of projects can cause performance issues',
```

src/renderer/components/EnhancedTable/EnhancedTableRow.js
```diff
@@ -82,6 +83,50 @@ const EnhMore = styled.div`
...
+
+  &:hover {
+    div {
+      opacity: 1;
+      visibility: visible;
+    }
+  }
+`;
+
+const WarningTooltip = styled.div`
```
stefcameron, line 98: As discussed on Slack, we'll use Lens' Tooltip component.

stefcameron (changes requested):

stefcameron (approved):
Looks good, thanks!

PR #837, opened 2022-04-21
[PRODX-23450]: fixed bug with checkboxes state

AntonNos:


AntonNos:
> What I see now is that if the "all" checkbox is indeterminate/partially checked and I click it, it gets unchecked (and unchecks all the children). It should be the opposite: If the "all" checkbox is partially checked and then clicked, it should go to fully checked (and all children checked).
> 
> The other issues are fixed, though! 🎉

Fixed, check please

stefcameron (approved):
This is working well!

PR #868, opened 2022-05-16
[PRODX-22916]: Added onSuspend and onResume events

AntonNos:
Adds power suspend/resume and network offline/online detection.

Note that network offline/online detection is not 100% reliable since Lens could be running
in a VM with its network adapter configured to be "always connected", which would result
in false-negatives if the host loses network connectivity (i.e. host goes offline, and we think
we're still connected because we don't receive an "offline" event).

---

I also realized, that we can debug in console IpcMain events. We can see network statuses in console.
Screenshot 2022-05-16 at 14 15 36

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -36,47 +36,39 @@ export default class ExtensionMain extends Main.LensExtension {
     },
   ];
 
+  powerDisposers = [];
+
   onActivate() {
     logger.log('ExtensionMain.onActivate()', 'extension activated');
-
-    const ipc = IpcMain.createInstance(this);
 
     // NOTE: an extension can only have ONE registered Catalog Source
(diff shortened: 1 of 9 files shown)


src/main/main.ts
```diff
@@ -36,22 +36,39 @@ export default class ExtensionMain extends Main.LensExtension {
...
+      ipcMain: ipc,
+    });
+
+    this.powerDisposers.push(
+      Main.Power.onSuspend(async () => ipc.notifyPowerSuspend())
+    );
+
+    this.powerDisposers.push(
+      Main.Power.onResume(async () => ipc.notifyPowerResume())
+    );
```
stefcameron, line 65: @AntonNos You were close... This is the most important bit you were missing. I also added network online/offline detection, which is done in `renderer.tsx`.

PR #876, opened 2022-05-18
[PRODX-24070]: Leverage new Minimum Viable Value feature in RTV.js to reduce memory use

AntonNos:


stefcameron (approved):
Looks great! 😄

PR #887, opened 2022-05-25
[PRODX-24176]: added tests for TriStateCheckbox component

AntonNos:
Screenshot 2022-05-25 at 13 35 57

### DIFF
--- a/src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
+++ b/src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
@@ -77,7 +77,6 @@ export const TriStateCheckbox = ({ label, onChange, value }) => {
         </CheckboxControl>
         <CheckboxLabel>
           <CheckboxField
-            data-testid="checkbox-field"
             type="checkbox"
             checked={value !== checkValues.UNCHECKED}
             onChange={onChange}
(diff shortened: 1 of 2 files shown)


src/renderer/components/TriStateCheckbox/TriStateCheckbox.js
```diff
@@ -77,6 +77,7 @@ export const TriStateCheckbox = ({ label, onChange, value }) => {
         </CheckboxControl>
         <CheckboxLabel>
           <CheckboxField
+            data-testid="checkbox-field"
```
stefcameron, line 80: Is it really necessary to use a test ID? In this case, you've put it on what is essentially the `<input type="checkbox">` element. You can find this with `screen.getByRole('checkbox')` or if you know the Checkbox's label, you can use `screen.getByLabelText(label)`

https://testing-library.com/docs/queries/byrole
https://testing-library.com/docs/queries/bylabeltext

I've also added a custom query `screen.getBySelector('input[type="checkbox"]')` (also findBySelector(), queryBySelector(), etc.) as you can see here: https://github.com/Mirantis/lens-extension-cc/blob/master/tools/tests/customQueries.js

AntonNos: Thanks for comment, I will take into account, that test IDs is bad practice in our case. 
Fixed

stefcameron (approved):
Looks good!

PR #921, opened 2022-06-14
[PRODX-24170]: added tests for WelcomeView component

AntonNos:
Screenshot 2022-06-14 at 13 39 21

### DIFF
--- a/__mocks__/@k8slens/extensions.js
+++ b/__mocks__/@k8slens/extensions.js
@@ -1,6 +1,7 @@
 import { computed, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
 import React from 'react';
+import propTypes from 'prop-types';
 import ReactSelect, { components } from 'react-select';
 import ReactSelectCreatable from 'react-select/creatable';
 
@@ -180,6 +181,19 @@ export const Common = {
   },
 };
 
+const Button = ({ label, primary, ...props }) => {
+  return (
+    <button className={primary ? 'primary' : ''} {...props}>
+      {label}
+    </button>
+  );
+};
+
+Button.propTypes = {
+  label: propTypes.string,
+  primary: propTypes.bool,
+};
+
 export const Renderer = {
   Ipc,
   Component: {
@@ -187,11 +201,7 @@ export const Renderer = {
     Notifications: {
       error: () => {},
     },
-    Button: (props) => {
-      // render label as text so that we can use .getByText
-      // eslint-disable-next-line react/prop-types
-      return <button disabled={props.disabled}>{props.label}</button>;
-    },
+    Button: Button,
     Icon: (props) => <i {...props} />,
     ConfirmDialog: {
       // eslint-disable-next-line no-undef
--- a/src/renderer/components/GlobalPage/WelcomeView.js
+++ b/src/renderer/components/GlobalPage/WelcomeView.js
@@ -108,6 +108,8 @@ const WelcomeButtonWrapper = styled.div(function () {
   };
 });
 
+const welcomeListItems = strings.welcome.listItems();
+
 export const WelcomeView = ({ openAddCloud }) => (
   <WelcomeContainer>
     <WelcomeWrapper>
@@ -121,11 +123,11 @@ export const WelcomeView = ({ openAddCloud }) => (
           }}
         />
         <WelcomeDescription>{strings.welcome.description()}</WelcomeDescription>
-        <WelcomeList
-          dangerouslySetInnerHTML={{
-            __html: strings.welcome.listItemsHtml(),
-          }}
-        />
+        <WelcomeList>
+          {welcomeListItems.map((item, index) => (
+            <li key={index}>{item}</li>
+          ))}
+        </WelcomeList>
         <WelcomeLink href={constants.repository.readmeUrl} target="_blank">
           {strings.welcome.link.label()}
         </WelcomeLink>
--- a/src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
+++ b/src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
@@ -1,0 +1,62 @@
+import { render, screen } from 'testingUtility';
+import userEvent from '@testing-library/user-event';
+import { WelcomeView } from '../WelcomeView';
+import * as strings from '../../../../strings';
+import * as constants from '../../../../constants';
+
+describe('/renderer/components/GlobalPage/WelcomeView', () => {
+  let user;
+
+  beforeEach(() => {
+    user = userEvent.setup();
+  });
+
+  it('renders an Welcome page', () => {
+    render(<WelcomeView openAddCloud={() => {}} />);
+
+    const removeHtmlTags = (string) => {
+      return string.replace(/<[^>]*>/g, '');
+    };
+
+    const containerCloudIcon = screen.getBySelector('svg');
+    const containerCloudIconSize = '80';
+    const containerCloudIconColor = 'var(--textColorAccent)';
+
+    const welcomeLink = screen.getByText(strings.welcome.link.label());
+
+    expect(containerCloudIcon).toHaveAttribute('width', containerCloudIconSize);
+    expect(containerCloudIcon).toHaveAttribute(
+      'height',
+      containerCloudIconSize
+    );
+    expect(containerCloudIcon).toHaveStyle(`fill: ${containerCloudIconColor}`);
+
+    expect(
+      screen.getByText(removeHtmlTags(strings.welcome.titleHtml()))
+    ).toBeInTheDocument();
+
+    expect(screen.getByText(strings.welcome.description())).toBeInTheDocument();
+
+    expect(screen.getBySelector('ul')).toHaveTextContent(
+      strings.welcome.listItems().join('')
+    );
+
+    expect(welcomeLink).toBeInTheDocument();
+    expect(welcomeLink).toHaveAttribute('target', '_blank');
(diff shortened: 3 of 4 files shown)


src/strings.ts
```diff
@@ -349,11 +349,11 @@ export const welcome: Dict = {
...
-  listItemsHtml: () => `
-    <li>Synchronizing projects to view their resources in the Lens Catalog</li>
-    <li>Monitoring real-time cluster status and resource utilization in Lens</li>
-    <li>Filtering synced resources and viewing their details in the Lens Catalog</li>
-  `,
+  listItems: () => [
+    'Synchronizing projects to view and manage their resources in the Lens Catalog',
+    'Monitoring real-time cluster status and resource utilization in Lens',
+    'Creating and deleting clusters in your projects from Lens',
+  ],
```
stefcameron, line 356: It's OK to remove the HTML markup, but you're using the wrong content. Please use the content from the `master` branch and just remove the markup.

src/renderer/components/GlobalPage/WelcomeView.js
```diff
@@ -121,17 +123,17 @@ export const WelcomeView = ({ openAddCloud }) => (
...
+            <li key={index}>{item}</li>
+          ))}
+        </WelcomeList>
         <WelcomeLink href={constants.repository.readmeUrl} target="_blank">
           {strings.welcome.link.label()}
         </WelcomeLink>
         <WelcomeButtonWrapper>
           <Component.Button
-            primary
+            primary="true"
```
stefcameron, line 136: This looks like a change to the code to accommodate the test, which is not a good practice.

Though looking at your tests, I don't think you're actually looking for a `button` element that has a `primary` attribute set to `"true"` anywhere, so you can probably just revert this change back to `primary`.

But if I missed something that you're indeed looking for a "primary button", not just a "button", then you should update the Button component's mock to specially-handle a boolean prop named `primary` and do with it what the real Lens component would normally do with it. This is key, because if you just do whatever to get the test to pass, the test is unreliable because it gives us a false-positive (i.e. we think we didn't break anything because the tests passed, when in reality, we may have broken something).

src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    expect(welcomeList).toHaveTextContent(strings.welcome.listItems().join(''));
+
+    expect(welcomeLink).toHaveAttribute('target', '_blank');
+    expect(welcomeLink).toHaveAttribute('href', constants.repository.readmeUrl);
+    expect(welcomeLink).toHaveTextContent(strings.welcome.link.label());
+
+    expect(welcomeButton).toHaveTextContent(strings.welcome.button.label());
+  });
+
+  it('clicking on button triggers onClick handler', async () => {
```
stefcameron, line 61: This may seem like splitting hairs (i.e. being too particular), but I'm just trying to help you look at tests with the right perspective, so I hope this help.

"clicking on button triggers onClick handler" isn't really relevant to the WelcomePage component. What you want to test is, "clicking on button triggers openAddCloud handler" -- you don't really know (i.e. this component doesn't really know) __how__ that handler is being triggered (could be a click, could be mouse down, could be hover, could be some other event); you just care that it's __being__ triggered.

It's a subtle difference, but it stresses the principal of "separation of concerns", which will serve you extremely well in any code you write. It applies to nearly all things we do as software engineers when we write code or design systems. 😄 

```suggestion
  it('clicking on button triggers openAddCloud handler', async () => {
```

src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
```diff
@@ -0,0 +1,70 @@
...
+
+    expect(welcomeDescription).toHaveTextContent(strings.welcome.description());
+
+    expect(welcomeList).toHaveTextContent(strings.welcome.listItems().join(''));
+
+    expect(welcomeLink).toHaveAttribute('target', '_blank');
+    expect(welcomeLink).toHaveAttribute('href', constants.repository.readmeUrl);
+    expect(welcomeLink).toHaveTextContent(strings.welcome.link.label());
+
+    expect(welcomeButton).toHaveTextContent(strings.welcome.button.label());
```
stefcameron, line 58: This is nice, but since there's literally not one branch of code in the WelcomeView (i.e. there are no conditional statements), there isn't much value checking all these things. I bet you could remove this test and still be at 100% coverage just with the test that checks the `openAddCloud` handler is called when the button is clicked.

What this will do, however, is cause tests to break if there's the slightest change to the WelcomeView's implementation which would probably inconsequential, which will make tests harder to maintain.

But we can keep this. I'm just pointing out that your time is better spent testing branch logic than it is making sure the right strings are being used.

As for the button, it would be better to use `screen.getByLabelText(strings.welcome.button.label())` than `firstChild.querySelector('button')` to get the element because that would test all these things at once:

1. A button is being used
2. The button is using the expected label
3. There is a single button with that label on the screen.

When you use `firstChild.querySelector('button');`, you're just finding the first button element. There might be another, and even with the same label, but you'll totally miss that.

stefcameron (changes requested):
Nice work! All comments except one are more for educational purposes than being changes you should make. The one comment about the list item strings, however, needs to be addressed. 😄

src/strings.ts
AntonNos: Done

src/renderer/components/GlobalPage/WelcomeView.js
AntonNos: Updated mocks to to get rid of console error.

src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
AntonNos: Understand, thanks

src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
AntonNos: I changed `welcomeButton` to what you said, but received an error, that can't find element with label `strings.welcome.button.label()`.
So I used `getAllByText`, but it found 2 buttons. I updated PR, check please what need to be changed.
Also I understand, that I did a lot of unnecessary tests. Will try no to do them in the future.

__mocks__/@k8slens/extensions.js
```diff
@@ -211,3 +216,8 @@ export const Renderer = {
     KubeObject: class {},
   },
 };
+
+Button.propTypes = {
+  label: propTypes.string,
+  primary: propTypes.bool,
+};
```
stefcameron, line 223: Since this is directly related to the definition of the mocked Button component on lines 184-190, could we move this up to line 192 and move `export const Renderer = {...}` down? That would make the code more readable.

src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const welcomeTitle = firstChild.querySelector('h1');
+
+    const welcomeDescription = firstChild.querySelector('p');
+
+    const welcomeList = firstChild.querySelector('ul');
+
+    const welcomeLink = firstChild.querySelector('a');
+
+    const welcomeButton = screen.getAllByText(strings.welcome.button.label());
```
stefcameron, line 37: This is strange. Why wouldn't `getByLabelText(...)` work here? Also, when combined with `.toBeInTheDocument()`, it's best to use `queryBy...()` queries instead of `getBy` or `getAllBy` because the "get" queries __require__ finding what they're looking for and will throw if not found (so the existence is tied with the query), whereas "query" queries don't require existence (i.e. they can return `null` for `queryBy` and an empty list for `queryAllBy`). You can read about the different types of queries here: https://testing-library.com/docs/queries/about#types-of-queries

So in the end, lines 37 and 60 should come down to this:

```javascript
expect(screen.queryByLabelText(strings.welcome.button.label())).toBeInTheDocument();
```

If for some reason the button doesn't exist, the `queryBy` query will return `null` (instead of `getBy` which would throw), and then the `toBeInTheDocument` assertion will fail the test because `null` indicates it's _not_ in the document.

Otherwise, `queryBy` will return the button, and the assertion will pass.

stefcameron (changes requested):
Let's see if we can address that `getAllByText()` issue... 🙂

__mocks__/@k8slens/extensions.js
AntonNos: Sure, let me check all your comments and I will update PR

src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+  beforeEach(() => {
+    user = userEvent.setup();
+  });
+
+  it('renders an Welcome page', () => {
+    render(<WelcomeView openAddCloud={() => {}} />);
+
+    const {
+      container: { firstChild },
```
stefcameron, line 18: Also, as I mentioned on other PRs, this isn't a good practice. Prefer using `screen` and `document` instead of doing this.

src/renderer/components/GlobalPage/__tests__/WelcomeView.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    } = render(<WelcomeView openAddCloud={() => {}} />);
```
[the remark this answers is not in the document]
AntonNos: Thanks! Fixed

stefcameron (approved):

PR #929, opened 2022-06-20
[PRODX-24171]: Added tests for ErrorPanel and InfoPanel components

AntonNos:
Screenshot 2022-06-20 at 14 04 07

### DIFF
--- a/src/renderer/components/__tests__/ErrorPanel.spec.js
+++ b/src/renderer/components/__tests__/ErrorPanel.spec.js
@@ -1,0 +1,24 @@
+import { render, screen } from 'testingUtility';
+import { ErrorPanel } from '../ErrorPanel';
+
+describe('/renderer/components/ErrorPanel', () => {
+  const message = 'test message';
+
+  it('renders |without| a message', () => {
+    render(<ErrorPanel />);
+
+    expect(
+      screen.getBySelector('i[material="error_outline"]')
+    ).toBeInTheDocument();
+    expect(screen.getBySelector('p')).toHaveTextContent('');
+  });
+
+  it('renders |with| a message', () => {
+    render(<ErrorPanel>{message}</ErrorPanel>);
+
+    expect(
+      screen.getBySelector('i[material="error_outline"]')
+    ).toBeInTheDocument();
+    expect(screen.getBySelector('p')).toHaveTextContent(message);
+  });
+});
--- a/src/renderer/components/__tests__/InfoPanel.spec.js
+++ b/src/renderer/components/__tests__/InfoPanel.spec.js
@@ -1,0 +1,24 @@
+import { render, screen } from 'testingUtility';
+import { InfoPanel } from '../InfoPanel';
+
+describe('/renderer/components/InfoPanel', () => {
+  const message = 'test message';
+
+  it('renders |without| a message', () => {
+    render(<InfoPanel />);
+
+    expect(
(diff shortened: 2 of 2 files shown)


src/renderer/components/__tests__/ErrorPanel.spec.js
```diff
@@ -0,0 +1,28 @@
...
+      container: { firstChild },
+    } = render(<ErrorPanel />);
+
+    expect(
+      firstChild.querySelector('i[material="error_outline"]')
+    ).toBeInTheDocument();
+    expect(firstChild.querySelector('p')).toHaveTextContent('');
+  });
+
+  it(`renders an Error Panel with |${randomString}| message`, () => {
```
stefcameron, line 18: Since the `randomString` is random, it's not useful having it in the test name like this. I would put the pipes around "with" and "without" between this and the test above it since that's the important difference between the two. So "renders |without| a message" and "renders |with| a message" would be good names for those tests. No need to repeat "Error Panel" because we know these tests are all related to the ErrorPanel from the `describe('/renderer/components/ErrorPanel', () => {` spec title.

src/renderer/components/__tests__/ErrorPanel.spec.js
```diff
@@ -0,0 +1,28 @@
+import { render } from 'testingUtility';
+import { ErrorPanel } from '../ErrorPanel';
+
+describe('/renderer/components/ErrorPanel', () => {
+  const randomString = (Math.random() + 1).toString(36).substring(7);
+
+  it('renders an Error Panel without message', () => {
+    const {
+      container: { firstChild },
```
stefcameron, line 9: Instead of always getting the first child like this, it's a better practice to use `screen`. When rendered, what you render should always be the only thing that's in the DOM anyway. And using `firstChild` like this might result in hiding multi-render issues with tests like on that other PR for the WelcomeView.

If you need to query for an element, you can just use `document.querySelector()` right there wherever you have `firstChild.querySelector()`.

src/renderer/components/__tests__/InfoPanel.spec.js
```diff
@@ -0,0 +1,28 @@
+import { render } from 'testingUtility';
+import { InfoPanel } from '../InfoPanel';
+
+describe('/renderer/components/InfoPanel', () => {
+  const randomString = (Math.random() + 1).toString(36).substring(7);
+
+  it('renders an Info Panel without message', () => {
+    const {
+      container: { firstChild },
```
stefcameron, line 9: Same here...

src/renderer/components/__tests__/InfoPanel.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    } = render(<InfoPanel />);
+
+    expect(
+      firstChild.querySelector('i[material="info_outline"]')
+    ).toBeInTheDocument();
+    expect(firstChild.querySelector('p')).toHaveTextContent('');
+  });
+
+  it(`renders an Info Panel with |${randomString}| message`, () => {
```
stefcameron, line 18: Same here...

stefcameron (changes requested):
Nice coverage!

src/renderer/components/__tests__/ErrorPanel.spec.js
AntonNos: Thanks, fixed all comments

stefcameron (approved):

PR #930, opened 2022-06-20
[PRODX-24171]: added tests for ContainerCloudIcon component

AntonNos:
Screenshot 2022-06-20 at 15 44 29


src/renderer/components/__tests__/ContainerCloudIcon.spec.js
```diff
@@ -0,0 +1,32 @@
+import { render } from 'testingUtility';
+import { ContainerCloudIcon } from '../ContainerCloudIcon';
+
+describe('/renderer/components/ContainerCloudIcon', () => {
+  it('renders an default Container Cloud icon', () => {
+    const {
+      container: { firstChild },
```
stefcameron, line 7: As I mentioned on other PRs, this isn't a good practice. Prefer using `screen` and `document` instead of doing this.

stefcameron (changes requested):
Nice!

src/renderer/components/__tests__/ContainerCloudIcon.spec.js
AntonNos: Fixed

stefcameron (changes requested):

stefcameron (approved):

PR #932, opened 2022-06-21
[PRODX-24171]: added tests for InlineNotice component

AntonNos:
Screenshot 2022-06-21 at 14 17 51

### DIFF
--- a/src/renderer/components/__tests__/InlineNotice.spec.js
+++ b/src/renderer/components/__tests__/InlineNotice.spec.js
@@ -1,0 +1,64 @@
+import { render, screen } from 'testingUtility';
+import { InlineNotice, types, iconSizes } from '../InlineNotice';
+
+describe('/renderer/components/InlineNotice', () => {
+  it('renders an default notice', () => {
+    render(<InlineNotice />);
+
+    expect(screen.getBySelector('i[material="info"]')).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


src/renderer/components/__tests__/InlineNotice.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    expect(firstChild.querySelector('i[material="info"]')).toBeInTheDocument();
+    expect(firstChild.querySelector('i[material="info"]')).toHaveStyle(
+      'color: var(--colorInfo)'
+    );
+  });
+
+  Object.values(types).forEach((type) => {
+    Object.values(iconSizes).forEach((size) => {
+      it(`renders an InlineNotice component with |${type}| icon, |${size}| size and custom content`, () => {
+        const randomString = (Math.random() + 1).toString(36).substring(7);
```
stefcameron, line 19: Looking at this again, which you've used in all your other PRs, I'm not so sure this is the best idea, because it's random. Tests should be reproducible, especially when they fail, so they're relatively easy to debug. I would recommend not doing this and just declaring `const text = 'text message'`, for example. Do the same for all the other PRs where you use a random string. It will remove variability, which will make tests more predictable.

stefcameron (changes requested):

stefcameron (changes requested):

PR #933, opened 2022-06-21
[PRODX-24171]: added tests for Section component

AntonNos:
Screenshot 2022-06-21 at 14 38 31


src/renderer/components/__tests__/Section.spec.js
```diff
@@ -0,0 +1,12 @@
+import { render } from 'testingUtility';
+import { Section } from '../Section';
+
+describe('/renderer/components/Section', () => {
+  it('renders an Section', () => {
+    const {
+      container: { firstChild },
```
stefcameron, line 7: As I mentioned on other PRs, this isn't a good practice. Prefer using `screen` and `document` instead of doing this.

src/renderer/components/__tests__/Section.spec.js
```diff
@@ -0,0 +1,12 @@
...
+    } = render(<Section />);
+
+    expect(firstChild).toBeInTheDocument();
```
stefcameron, line 10: Let's also add a check that the rendered tag is 'SECTION', which would be solved in a single expectation:
```
expect(screen.getBySelector('section')).toBeInTheDocument();
```

stefcameron (changes requested):

src/renderer/components/__tests__/Section.spec.js
AntonNos: Done

stefcameron (approved):

PR #934, opened 2022-06-21
[PRODX-24171]: added tests for styles

AntonNos:
Screenshot 2022-06-21 at 15 34 38

### DIFF
--- a/src/renderer/components/__tests__/styles.spec.js
+++ b/src/renderer/components/__tests__/styles.spec.js
@@ -1,0 +1,23 @@
+import * as styles from '../styles';
+
+describe('/renderer/components/styles', () => {
+  describe('mixinFlexColumnGaps()', () => {
+    it('return styles that define a flex container, column-oriented, with default consistent gaps between its _immediate_ children', () => {
+      expect(styles.mixinFlexColumnGaps()['> *'].marginBottom).toBe(
+        styles.layout.gap
+      );
+    });
+
+    it('return styles that define a flex container, column-oriented, with custom consistent gaps between its _immediate_ children', () => {
+      expect(styles.mixinFlexColumnGaps(10)['> *'].marginBottom).toBe(10);
+    });
+  });
+
+  describe('mixinColumnStyles()', () => {
+    it('return styles for columns in a Flex column layout', () => {
+      expect(styles.mixinColumnStyles()['> *'].marginBottom).toBe(
+        styles.layout.grid * 6
+      );
+    });
+  });
+});


src/renderer/components/__tests__/styles.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    it('return common styles for pages', () => {
+      expect(styles.mixinPageStyles()).toEqual({
+        padding: 16,
+        backgroundColor: 'var(--mainBackground)',
+        code: {
+          fontSize: 'calc(var(--font-size) * .9)',
+        },
+      });
+    });
+  });
```
stefcameron, line 59: I don't think this is worth testing. There's nothing parameterized here, just straight object export. Having a test for this will just make it more work to update the styles.

src/renderer/components/__tests__/styles.spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+  describe('mixinColumnStyles()', () => {
+    it('return styles for columns in a Flex column layout', () => {
+      expect(styles.mixinColumnStyles()).toEqual({
+        flex: 1,
+        display: 'flex',
+        flexDirection: 'column',
+        '> *': { marginBottom: 24 },
```
stefcameron, line 38: I think this (line 38) is the only thing that's worth testing here, the expectation that the margin is set to 6 x gap. And I would do it like this instead of hard coding numbers:

```
expect(styles.mixinColumnStyles()['> *'].marginBottom).toBe(styles.layout.grid * 6)
```

That way, if the value of `grid` is changed, the test will survive the change instead of breaking. It'll only break if the gap is made larger/smaller, and maybe that's an interesting double-check to make sure that's what we really meant to do.
```

File: src/renderer/components/__tests__/styles.spec.js [hunk 3]
```diff
@@ -0,0 +1,60 @@
+import * as styles from '../styles';
+
+describe('/renderer/components/styles', () => {
+  describe('mixinFlexColumnGaps()', () => {
+    it('return styles that define a flex container, column-oriented, with default consistent gaps between its _immediate_ children', () => {
+      expect(styles.mixinFlexColumnGaps()).toEqual({
+        display: 'flex',
+        flexDirection: 'column',
+        '> *': {
+          marginBottom: 16,
```
stefcameron, line 10: I would check this as `marginBottom: styles.layout.gap` in case `styles.layout.grid` changes, making `styles.layout.gap` change and resulting in the test breaking. We should just care here that the bottom margin defaults to `styles.layout.gap`.

I'd also not test the rest of the object since it's not really that important.

```
expect(styles.mixinFlexColumnGaps()['> *'].marginBottom).toBe(styles.layout.gap)
```

src/renderer/components/__tests__/styles.spec.js
```diff
@@ -0,0 +1,60 @@
...
+        },
+      });
+    });
+
+    it('return styles that define a flex container, column-oriented, with custom consistent gaps between its _immediate_ children', () => {
+      expect(styles.mixinFlexColumnGaps(10)).toEqual({
+        display: 'flex',
+        flexDirection: 'column',
+        '> *': {
+          marginBottom: 10,
```
stefcameron, line 23: Again with this, what's most interesting is that the value given as a parameter is used inside as expected, so

```
expect(styles.mixinFlexColumnGaps()['> *'].marginBottom).toBe(10)
```

would be enough for this.

stefcameron (changes requested):

src/renderer/components/__tests__/styles.spec.js
AntonNos: Fixed all comments

stefcameron (changes requested):

PR #939, opened 2022-06-27
[PRODX-24171]: added tests for useCheckboxes hooks

AntonNos:
UPD: 90+% coverage. Screenshot added.
Screenshot 2022-06-28 at 15 27 11


src/renderer/components/hooks/__tests__/useCheckboxes.spec.js
```diff
@@ -0,0 +1,199 @@
...
+            syncedNamespaces.push(name);
+          } else {
+            ignoredNamespaces.push(name);
+          }
+        });
+
+        expect(screen.getByText(syncedNamespaces.join(''))).toBeInTheDocument();
+        expect(
+          screen.getByText(ignoredNamespaces.join(''))
+        ).toBeInTheDocument();
```
stefcameron, line 142: I think these expectations would be stronger if you were more explicit about which text you were comparing to.

In your `TestComponent`, I would do this:
```javascript
          return (
            <div>
              <p data-testid="synced">{syncedNamespaces}</p>
              <p data-testid="ignored">{ignoredNamespaces}</p>
            </div>
          );
```

And here, I would do this:
```suggestion
        expect(screen.getByTestId('synced')).toHaveTextContent(syncedNamespaces.join(''));
        expect(
          screen.getByTestId('ignored')).toHaveTextContent(ignoredNamespaces.join(''));
```

src/renderer/components/hooks/__tests__/useCheckboxes.spec.js
```diff
@@ -0,0 +1,199 @@
...
+        const { getCheckboxValue } = useCheckboxes.useCheckboxes(
+          useCheckboxes.makeCheckboxesInitialState(testData)
+        );
+
+        return (
+          <div>
+            <label>
+              <input
+                type="checkbox"
+                checked={getCheckboxValue({ isParent: true }) !== 'UNCHECKED'}
```
stefcameron, line 157: Use the predefined constant for `'UNCHECKED'` instead of hard-coding the value.

stefcameron (approved):
Nice work, and some elaborate tests, too! Way to go! 🎉 👍

PR #956, opened 2022-07-05
WIP [PRODX-24169]: added tests to EnhancedTableHead

AntonNos:


src/renderer/components/EnhancedTable/__tests__/EnhancedTableHead.spec.js
```diff
@@ -0,0 +1,69 @@
...
+describe('/renderer/components/EnhancedTable/EnhancedTableHead', () => {
+  const testString = 'test';
+
+  let user;
+
+  beforeEach(() => {
+    user = userEvent.setup();
+  });
+
+  [true, false].forEach((bool) => {
```
stefcameron, line 16: Could you use a better name than `bool`? It doesn't tell me what it's meant to represent so I can tell what the difference is between true and false.

stefcameron (commented):
This PR is still WIP but it seems good so far!

src/renderer/components/EnhancedTable/__tests__/EnhancedTableHead.spec.js
AntonNos: Fixed

stefcameron (approved):
👍

PR #958, opened 2022-07-06
[PRODX-24169]: added tests to AdditionalInfoRows component

AntonNos:
Screenshot 2022-07-06 at 15 55 42
There is 100% coverage, but need to fix `console.error` before merging

UPD: error fixed

### DIFF
--- a/__mocks__/@k8slens/extensions.js
+++ b/__mocks__/@k8slens/extensions.js
@@ -136,6 +136,18 @@ class KubernetesCluster {
   }
 }
 
+class CatalogEntity {
+  constructor(props) {
+    Object.assign(this, props);
+  }
+}
+
+class CatalogCategory {
+  constructor(props) {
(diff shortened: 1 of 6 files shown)


__mocks__/@k8slens/extensions.js
```diff
@@ -98,19 +98,86 @@ export class _Select extends React.Component {
 
 export const Select = observer(_Select);
 
+// class Singleton {
```
stefcameron, line 101: I'm curious why it looks like you essentially copy/pasted Lens' Singleton implementation into our mocks. Was there something specific in there you were needing?

AntonNos: Hmm, you are right. Idk why this code was here, maybe I tested something and forgot to remove it.
Removed it, thank you

stefcameron (approved):
Looks good!

PR #961, opened 2022-07-11
WIP [PRODX-24169]: added tests to EnhancedTable component

AntonNos:


stefcameron:
👀

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
```diff
@@ -0,0 +1,166 @@
...
+          proxyCount: 0,
+          sshKeyCount: 2,
+          synced: true,
+        },
+      ],
+    });
+  });
+
+  [true, false].forEach((isSelectiveSyncView) => {
+    it(`renders a table when isSelectiveSyncView is |${isSelectiveSyncView}|`, () => {
```
stefcameron, line 87:
```suggestion
    it(`|${isSelectiveSyncView ? 'does not render' : 'renders'}| a table when isSelectiveSyncView is |${isSelectiveSyncView}|`, () => {
```

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
```diff
@@ -0,0 +1,166 @@
...
+          isSyncStarted={false}
+        />
+      </CloudProvider>
+    );
+
+    const expectedAscCloudsName = ['foo', 'bar'];
+    const expectedDescCloudsName = ['bar', 'foo'];
+
+    // Clouds should be ordered by |asc| (first render)
+    // screen.debug();
```
stefcameron, line 120: Remove this comment please! 🙂 

Tip: Whenever I add or comment code for debugging purposes, I always associate a `// DEBUG` comment with it because our lint rules are such that those types of comments will cause the linter to fail.

So in this case, I would've started with
```
screen.debug(); // DEBUG
```

And then either removed the whole line, or if commenting, would've commented like this:
```
// DEBUG screen.debug();
```

That way, the PR would've failed the `yarn lint` test and you would've remembered to remove this line. 🙂

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
```diff
@@ -0,0 +1,166 @@
...
+      <CloudProvider>
+        <EnhancedTable
+          clouds={{ 'foo.com': fakeCloudFoo, 'bar.com': fakeCloudBar }}
+          isSyncStarted={false}
+        />
+      </CloudProvider>
+    );
+
+    const expectedAscCloudsName = ['foo', 'bar'];
+    const expectedDescCloudsName = ['bar', 'foo'];
```
stefcameron, line 117: Why does this order seem to be the opposite of what it should be? In ascending order, shouldn't it be "bar, foo", and then in descending order, "foo, bar"? Alphabetically, ascending is a to z, so bar then foo... 🤔

stefcameron (changes requested):
Nice, though I have a question...

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
AntonNos: Yes, sorry, I missed this one. Removed

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
AntonNos: You are right, fixed

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
```diff
@@ -0,0 +1,181 @@
...
+
+    const thirdTimeSortedByNameCloudsName = [];
+    document.querySelectorAll('i[material="chevron_right"]').forEach((icon) => {
+      thirdTimeSortedByNameCloudsName.push(
+        icon.parentNode.parentNode.textContent
+      );
+    });
+    expect(thirdTimeSortedByNameCloudsName).toEqual(expectedAscCloudsName);
+
+    // Clouds should be ordered by |asc|
```
stefcameron, line 168:
```suggestion
   // Clouds should be ordered by |desc|
```

stefcameron (approved):
Looks good!

PR #963, opened 2022-07-12
[PRODX-24169]: added tests for tableUtil

AntonNos:
Screenshot 2022-07-12 at 16 46 22

### DIFF
--- a/src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
+++ b/src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
@@ -1,0 +1,112 @@
+import * as tableUtil from '../tableUtil';
+
+describe('/renderer/components/EnhancedTable/tableUtil', () => {
+  describe('getTableData()', () => {
+    [true, false].forEach((isSelectiveSyncView) => {
+      it(`uses |${
+        isSelectiveSyncView
+          ? tableUtil.SELECTIVE_HEAD_CELL_VALUES
+          : tableUtil.HEAD_CELL_VALUES
+      }| as the head cell value when isSelectiveSyncView is |${isSelectiveSyncView}|`, () => {
+        if (isSelectiveSyncView) {
+          expect(tableUtil.getTableData(isSelectiveSyncView)).toMatchObject({
+            path: tableUtil.selectivePathToData,
+            headCellValue: tableUtil.SELECTIVE_HEAD_CELL_VALUES,
+          });
+        } else {
+          expect(tableUtil.getTableData(isSelectiveSyncView)).toMatchObject({
+            path: tableUtil.pathToData,
+            headCellValue: tableUtil.HEAD_CELL_VALUES,
+          });
+        }
+      });
+    });
+  });
+
+  describe('sortData()', () => {
+    const fakeClouds = {
+      'http://foo.com': {
+        name: 'foo',
+      },
+      'http://bar.com': {
+        name: 'bar',
+      },
+    };
+
+    it('returns sorted object keys array when order is |asc|', () => {
+      expect(
+        tableUtil.sortData(
+          fakeClouds,
+          tableUtil.HEAD_CELL_VALUES.NAME,
+          'asc',
+          tableUtil.selectivePathToData
+        )
+      ).toEqual([['http://foo.com'], ['http://bar.com']]);
+    });
+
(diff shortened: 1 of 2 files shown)


src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    USERNAME: 'USERNAME',
+    STATUS: 'STATUS',
+  };
+
+  const pathToData = {
+    [HEAD_CELL_VALUES.NAME]: ['name'],
+    [HEAD_CELL_VALUES.URL]: ['cloudUrl'],
+    [HEAD_CELL_VALUES.USERNAME]: ['username'],
+    [HEAD_CELL_VALUES.STATUS]: ['status'],
+  };
```
stefcameron, line 28: I think this stuff is all defined in tableUtil.js, right? Just export them and then import them into this spec file instead of duplicating them here. There isn't much point "testing" constants, and if their values change in the code, then the tests won't break so easily.

src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
```diff
@@ -0,0 +1,134 @@
...
+
+  describe('getTableData()', () => {
+    [true, false].forEach((isSelectiveSyncView) => {
+      it(`returns an object depends on |isSelectiveSyncView| value. |isSelectiveSyncView = ${isSelectiveSyncView}|`, () => {
```
stefcameron, line 32: I like that you've noticed my use of using the pipe (|) for the parameterized parts of a test in the descriptions, but try to keep them only around what changes. In this case,

```suggestion
      it(`uses |${isSelectiveSyncView ? SELECTIVE_HEAD_CELL_VALUES : HEAD_CELL_VALUES}| as the head cell value when isSelectiveSyncView is |${isSelectiveSyncView}|`, () => {
```

Or something similar. Just make it specific to what's expected instead of too generic.

src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    const testNamespacesWithDefault = [
+      {
+        name: 'test 1',
+      },
+      {
+        name: 'default',
+      },
+    ];
+
+    it('returns an initial object, if there is no |default| name or |default| is first', () => {
```
stefcameron, line 119: Here, "|default|" doesn't appear to be parameterized, so it should not be between pipes because it never changes (AFAICT).

src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
```diff
@@ -0,0 +1,134 @@
...
+      expect(
+        tableUtil.sortNamespaces(testNamespacesWithoutDefault)
+      ).toMatchObject(testNamespacesWithoutDefault.sort(compareNamespaces));
+      expect(
+        tableUtil.sortNamespaces(testNamespacesWithFirstDefault)
+      ).toMatchObject(testNamespacesWithFirstDefault.sort(compareNamespaces));
+    });
+
+    it('returns a sorted object, if there is not first |default| name', () => {
```
stefcameron, line 128: Same here...

stefcameron (changes requested):
Nice work on the coverage! 😄

src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
AntonNos: Fixed

src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
AntonNos: Got it, noted. Thanks.
Fixed

src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
AntonNos: Fixed

src/renderer/components/EnhancedTable/__tests__/tableUtil.spec.js
AntonNos: Fixed

stefcameron (approved):
👌

PR #975, opened 2022-07-18
[PRODX-24169]: added tests for EnhancedTableRow component

AntonNos:
Screenshot 2022-07-26 at 14 42 08


stefcameron:
👀

__mocks__/@k8slens/extensions.js
```diff
@@ -319,10 +391,14 @@ export const Common = {
...
-    getAppVersion: () => '1.0.0',
-  },
+  // Util: {
+  //   Singleton,
+  //   getAppVersion: () => '1.0.0',
+  //   openExternal: (url) => {
+  //     console.log(url);
+  //   }
+  // },
+  Util,
```
stefcameron, line 401: Looks like you've changed the Common.Util module to be a class that renders a URL. That is not a good idea. It might work for now, but it will break other tests down the road. We need to keep the mock definition of Common.Util as it was. If you didn't like that `openExternal(url)` was looking to the console, you can replace `console.log()` with code that creates a new element and inserts it into the DOM as a way of checking what URL was opened, but I would stick with the console statement. That can easily be checked in a test with jest-mock-console.

src/common/__tests__/MockCloud.js
```diff
@@ -2,30 +2,43 @@ import { EventDispatcher } from '../EventDispatcher';
 import { CloudNamespace } from '../CloudNamespace';
```
stefcameron, line 2: MockCloud.js should have been deleted...

src/renderer/components/EnhancedTable/__tests__/AdditionalInfoRows.spec.js
```diff
@@ -1,17 +1,20 @@
...
+
+jest.mock('../../../../common/Cloud');
 
 describe('/renderer/components/EnhancedRable/AdditionalInfoRows', () => {
   let fakeCloud;
+  let fakeCloudJson;
 
   beforeEach(() => {
-    fakeCloud = new Cloud('http://foo.com', {
+    fakeCloudJson = mkCloudJson({
```
stefcameron, line 12: Unless you need the JSON separately (i.e. because you need to initialize the CloudStore with it), you should just create the Cloud directly instead of having the JSON separate. That will keep the test file more concise and no one will wonder why this JSON variable exists. So just replace this with
```suggestion
    fakeCloud = new Cloud({
```

src/renderer/components/EnhancedTable/AdditionalInfoRows.js
```diff
@@ -68,20 +72,22 @@ export const AdditionalInfoRows = ({ namespace, emptyCellsCount }) => {
...
-              <EnhTableInfoRowCell isFirst>
-                {infoName} ({infoCount})
-              </EnhTableInfoRowCell>
-              {generateTableCells(emptyCellsCount)}
-            </EnhTableInfoRow>
-          ))}
-        </EnhInfoRowsBody>
-      </EnhInfoRowsTable>
-    </EnhInfoRowsWrapper>
+    <EnhInfoRow>
```
stefcameron, line 75: I'm curious what caused you to add the additional `<tr>` wrapping here. I don't see a change to the component that consumes the `<AdditionalInfoRows>` component, so I'm concerned that it's providing a `<tr>` and now we're ending up with `<tr><tr>...</tr></tr>` in the real code...

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
```diff
@@ -2,114 +2,93 @@ import mockConsole from 'jest-mock-console';
...
   beforeEach(() => {
     user = userEvent.setup();
     mockConsole(); // automatically restored after each test
 
     IpcRenderer.createInstance(extension);
-    CloudStore.createInstance().loadExtension(extension);
 
-    fakeCloudFoo = new Cloud('http://foo.com', {
-      name: 'foo',
+    fakeCloudFooJson = mkCloudJson({
```
stefcameron, line 26: Same comment here about `fakeCloudFooJson` and `fakeCloudBarJson` as on src/renderer/components/EnhancedTable/__tests__/AdditionalInfoRows.spec.js -- these don't appear to be needed independently from Cloud instances in order to initialize the CloudStore. Would be best to just not have them...

src/renderer/components/EnhancedTable/__tests__/EnhancedTableRow.spec.js
```diff
@@ -0,0 +1,401 @@
...
+    mockConsole(); // automatically restored after each test
+
+    IpcRenderer.createInstance(extension);
+  });
+
+  describe('render', () => {
+    let fakeCloudFoo,
+      fakeCloudFooJson,
+      fakeCloudWithoutNamespaces,
+      fakeCloudWithoutNamespacesJson;
```
stefcameron, line 44: Same comment here with `fakeCloudFooJson` and `fakeCloudWithoutNamespacesJson` as with src/renderer/components/EnhancedTable/__tests__/AdditionalInfoRows.spec.js -- these aren't needed in this test module to initialize the CloudStore, so best to just not have them.

src/renderer/components/EnhancedTable/__tests__/EnhancedTableRow.spec.js
```diff
@@ -0,0 +1,401 @@
+import mockConsole from 'jest-mock-console';
```
stefcameron, line 1: Nice work on the tests in here! 👍

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
```diff
@@ -102,21 +82,12 @@ describe('/renderer/components/EnhancedTable/TableRowListenerWrapper', () => {
...
-        proxyCount: 0,
-        sshKeyCount: 2,
-        synced: true,
-      },
-    ],
   });
 
+  const disconnectedFakeCloud = new Cloud(disconnectedFakeCloudJson);
+  const connectingFakeCloud = new Cloud(connectingFakeCloudJson);
+  const connectedFakeCloud = new Cloud(connectedFakeCloudJson);
```
stefcameron, line 89: All of these (and the related code above) should be in a `beforeEach()` to make sure that each test that uses them gets a clean set of data.

Also, again, as with src/renderer/components/EnhancedTable/__tests__/AdditionalInfoRows.spec.js, I don't think any of the separate `*FakeCloudJson` variables are needed in this test module. None of them appear to be used to initialize the CloudStore. Best not to have them.

stefcameron (changes requested):
Thanks for your work on the EnhancedTableRow tests, they look pretty good. Most comments are around the extensions.js mock changes, and updates to other tests related to the new mock Cloud class.

__mocks__/@k8slens/extensions.js
```diff
@@ -197,6 +197,70 @@ class Spinner extends React.Component {
...
+  }
+
+  static open = (props) => {
+    const { ok } = props;
+
+    confirmDialogInstance.setState({
+      isOpen: true,
+    });
+
+    ok();
```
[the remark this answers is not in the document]
AntonNos: Fixed

__mocks__/@k8slens/extensions.js
```diff
@@ -197,6 +197,70 @@ class Spinner extends React.Component {
...
+  };
+
+  render() {
+    return (
+      this.state.isOpen && (
+        <div className="confirm-buttons">
+          <button className="cancel"></button>
+          <button className="ok"></button>
```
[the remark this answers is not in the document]
AntonNos: Fixed

__mocks__/@k8slens/extensions.js
AntonNos: Fixed

src/common/__tests__/MockCloud.js
AntonNos: Done

src/renderer/components/EnhancedTable/__tests__/AdditionalInfoRows.spec.js
AntonNos: Done

src/renderer/components/EnhancedTable/AdditionalInfoRows.js
AntonNos: I added couple of `tr`, `td`, because there was an error with DOM validating. Something like `div` cannot be a child of `tr` etc..

src/renderer/components/EnhancedTable/__tests__/EnhancedTable.spec.js
AntonNos: Fixed

src/renderer/components/EnhancedTable/__tests__/EnhancedTableRow.spec.js
AntonNos: Done

src/renderer/components/EnhancedTable/__tests__/EnhancedTableRow.spec.js
AntonNos: Thanks!

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
AntonNos: I didn't put them in `beforeEach()`, because I have a loop: 
```
[disconnectedFakeCloud, connectingFakeCloud, connectedFakeCloud].forEach((cloud) => {
  it()
})
```.
And if I put them in `beforeEach(),` all of them are not defined --> `cloud` is `undefined`

**Comment by stefcameron** [comment]
👀

File: __mocks__/@k8slens/extensions.js [hunk 11]
```diff
@@ -197,6 +197,54 @@ class Spinner extends React.Component {
...
+    confirmDialogInstance = this;
+
+    this.state = {
+      isOpen: false,
+      ok: null,
+    };
+  }
+
+  static open = (props) => {
+    const { ok } = props;
```
stefcameron, line 218: Also handle `cancel` prop here; otherwise, when comes time to test the cancel for some future test, things will break and it'll require more time then to figure out why it's not working than it does now, knowing how it will fail... 🙂

__mocks__/@k8slens/extensions.js
```diff
@@ -197,6 +197,54 @@ class Spinner extends React.Component {
...
+
+    confirmDialogInstance.setState({
+      isOpen: true,
+      ok,
+    });
+  };
+
+  ok = () => {
+    this.state.ok();
```
stefcameron, line 227: Don't assume `ok` is defined. Use `this.state.ok?.()` (optional chaining; I think that's the syntax for an optional function call)

Also need to set `isOpen` state to false here too since OK also dismissed the modal.

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
```diff
@@ -153,7 +123,7 @@ describe('/renderer/components/EnhancedTable/TableRowListenerWrapper', () => {
   );
 
   it('renders a table row when cloud is fetching', () => {
-    const fetchingFakeCloud = new Cloud('http://barfoo.com', {
+    const fetchingFakeCloudJson = mkCloudJson({
```
stefcameron, line 126: You seem to have gone back to JSON from Cloud, but you're not using the JSON to initialize the CloudStore... Any reason why you went back to JSON?

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
```diff
@@ -212,12 +171,11 @@ describe('/renderer/components/EnhancedTable/TableRowListenerWrapper', () => {
           ? 'onCloudStatusChange()'
           : 'onCloudFetchingChange()'
       }| by dispatching |${event}| event`, async () => {
-        const fakeCloud = new Cloud('http://barfoo.com', {
+        const fakeCloudJson = mkCloudJson({
```
stefcameron, line 174: Same here?

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
```diff
@@ -276,12 +223,11 @@ describe('/renderer/components/EnhancedTable/TableRowListenerWrapper', () => {
   );
 
   it('triggers |onCloudSyncChange()| by dispatching |syncChange| event', async () => {
-    const fakeCloud = new Cloud('http://barfoo.com', {
+    const fakeCloudJson = mkCloudJson({
```
stefcameron, line 226: Same here?

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
```diff
@@ -102,19 +85,6 @@ describe('/renderer/components/EnhancedTable/TableRowListenerWrapper', () => {
...
-        machineCount: 12,
-        name: 'foobar',
-        proxyCount: 0,
-        sshKeyCount: 2,
-        synced: true,
-      },
-    ],
   });
 
   [disconnectedFakeCloud, connectingFakeCloud, connectedFakeCloud].forEach(
```
stefcameron, line 90: Since these clouds are really test cases as well as test data, make them consts and localize them to this particular test. I'm just concerned that having them in the outer scope will result in someone writing a test in the future using one of these, having the test modify it, without realizing that they aren't rebuilt with a `beforeEach()` on every test, and then we end-up with a really hard-to-find bug where one test depends on another's output.

So like this:

```suggestion
    (function () {
      const disconnectedFakeCloud = new Cloud(...);
      const connectingFakeCloud = new Cloud(...);
      const connectingFakeCloud = new Cloud(...);
      
      [disconnectedFakeCloud, connectingFakeCloud, connectedFakeCloud].forEach(
        ...
      );
    })();
```

stefcameron (changes requested):
Great updates! Just a few more loose ends to tie off.

__mocks__/@k8slens/extensions.js
AntonNos: Done

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
AntonNos: Hmm, maybe I missed that. Fixed

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
AntonNos: Fixed

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
AntonNos: Fixed

src/renderer/components/EnhancedTable/__tests__/TableRowListenerWrapper.spec.js
AntonNos: Hmm, good solution! Thank you, fixed

__mocks__/@k8slens/extensions.js
stefcameron: Thanks, but what you did is not what I was asking. 🙂 `cancel` is another prop like `ok` that the consumer can provide to the ConfirmDialog component and its `open()` method. The changes you made are still not handling this case. Handle the `cancel` prop the same way you're handling the `ok` prop.

AntonNos: Oh, I understand what you mean. Fixed

__mocks__/@k8slens/extensions.js
```diff
@@ -197,6 +197,62 @@ class Spinner extends React.Component {
...
+      ...this.state,
+      isOpen: false,
+    });
+  };
+
+  render() {
+    return (
+      this.state.isOpen && (
+        <div className="confirm-buttons">
+          <button className="cancel" onClick={this.state.cancel}></button>
```
stefcameron, line 248:
```suggestion
          <button className="cancel" onClick={this.cancel}></button>
```

__mocks__/@k8slens/extensions.js
```diff
@@ -197,6 +197,62 @@ class Spinner extends React.Component {
...
+      isOpen: true,
+      ok,
+      cancel,
+    });
+  };
+
+  ok = () => {
+    this.state.ok?.();
+    this.setState({
+      ...this.state,
```
stefcameron, line 231: Spreading the existing state into the state updater isn't necessary. React will take the object you give it here and shallow-merge it with the full existing state.

__mocks__/@k8slens/extensions.js
```diff
@@ -197,6 +197,62 @@ class Spinner extends React.Component {
...
+      isOpen: false,
+    });
+  };
+
+  cancel = () => {
+    this.state.cancel?.();
+    this.setState({
+      ...this.state,
```
stefcameron, line 239: Spreading the existing state into the state updater isn't necessary. React will take the object you give it here and shallow-merge it with the full existing state.

stefcameron (changes requested):

__mocks__/@k8slens/extensions.js
AntonNos: Missed it, sorry

stefcameron (approved):
👌

PR #982, opened 2022-07-19
[PRODX-24170]: added tests for SyncView component

AntonNos:
Reminder: Need to cover `handleAddCloud()` method
Screenshot 2022-07-20 at 14 23 21

UPD: this block wasn't covered after submitting. Should we cover it or we are good with this coverage percentage?

### DIFF
--- a/__mocks__/@k8slens/extensions.js
+++ b/__mocks__/@k8slens/extensions.js
@@ -252,9 +252,9 @@ export class ConfirmDialog extends React.Component {
   }
 }
 
-const Button = ({ label, primary, waiting, ...props }) => {
+const Button = ({ label, primary, waiting, plain, ...props }) => {
   return (
-    <button className={cx({ primary, waiting })} {...props}>
+    <button className={cx({ primary, waiting, plain })} {...props}>
       {label}
     </button>
   );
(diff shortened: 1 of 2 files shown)


__mocks__/@k8slens/extensions.js
```diff
@@ -264,15 +264,16 @@ Button.propTypes = {
...
   primary: propTypes.bool,
   waiting: propTypes.bool,
+  plain: propTypes.bool,
 };
 
-export const Icon = ({ interactive, smallest, ...props }) => {
-  return <i className={cx({ smallest, interactive })} {...props} />;
+const Input = ({ trim, ...props }) => {
+  const classes = [trim ? 'trim' : ''];
+  return <input className={classes.join()} {...props} />;
```
stefcameron, line 272: I think it'll be safer in the long term, as we add more classes for various scenarios, to use the `cx()` utility: `className={cx({ trim })}`

stefcameron (changes requested):
Just one small change.

__mocks__/@k8slens/extensions.js
AntonNos: Make sense. Fixed

stefcameron (approved):
Thank you!

stefcameron:
Merging now since I need to modify this new SyncView.spec.js file to add more tests for some local changes.

PR #983, opened 2022-07-19
[PRODX-24170]: added tests for GlobalPage component

AntonNos:
Screenshot 2022-08-02 at 00 15 49


src/renderer/components/GlobalPage/__tests__/GlobalPage.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      it('showing WelcomeView after rendering GlobalPage component', () => {
+        render(
+          <CloudProvider>
+            <GlobalPage />
+          </CloudProvider>
+        );
+
+        expect(
+          screen.getByText(strings.welcome.button.label())
+        ).toBeInTheDocument();
```
stefcameron, line 39: Instead of just checking it's in the document, let's also check that it's rendered using the light theme vs dark theme:
```suggestion
        expect(
          screen.getByText(strings.welcome.button.label())
        ).toHaveStyleRule('color', 'SOMETHING depending on light or dark mode');
```

stefcameron (changes requested):
Looks good. One suggestion for improvement.

src/renderer/components/GlobalPage/__tests__/GlobalPage.spec.js
AntonNos: I did something like ```
expect(screen.getByText(strings.welcome.description())).toHaveStyle(
          'color: var(--textColorPrimary);'
        );
```
But I can't check HEX color, because it told me, that received color in var

stefcameron (approved):

src/renderer/components/GlobalPage/__tests__/GlobalPage.spec.js
stefcameron: OK, not the best, but we'll stick with that. Thanks for updating the test!

PR #1000, opened 2022-08-01
Add Wizard component for use with Create Cluster wizard

stefcameron:
The Wizard is mostly modeled according to Lens' Preferences Panel,
but with additional control over the step list's functionality,
and back/next/finish buttons.

### Previews

__NOTE:__ The preview is not for the Create Cluster work, it's just for what the Wizard looks like.

Screen Shot 2022-08-01 at 5 46 06 PM

Screen Shot 2022-08-01 at 5 59 28 PM


AntonNos (approved):
Nice work on Wizard component! Everything looks great!

PR #1007, opened 2022-08-03
[PRODX-24170]: added tests for ConnectionBlock component

AntonNos:
Screenshot 2022-08-05 at 13 40 54

### DIFF
--- a/__mocks__/@k8slens/extensions.js
+++ b/__mocks__/@k8slens/extensions.js
@@ -1,6 +1,6 @@
 import { computed, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
-import React from 'react';
+import React, { useState } from 'react';
 import propTypes from 'prop-types';
 import ReactSelect, { components } from 'react-select';
 import ReactSelectCreatable from 'react-select/creatable';
@@ -258,6 +258,36 @@ class ConfirmDialog extends React.Component {
   }
 }
 
+const Input = ({ trim, ...props }) => {
+  const [errorMessage, setErrorMessage] = useState();
+  const [isValid, setIsValid] = useState(true);
+
+  const validate = (event) => {
+    props.validators.forEach((validator) => {
+      if (!validator.validate(event.target.value)) {
+        setIsValid(validator.validate(event.target.value));
+        setErrorMessage(validator.message());
+      }
+    });
+  };
+
+  const handleChange = (event) => {
+    props?.onChange(event.target.value, event);
+    validate(event);
+  };
+
+  return (
+    <>
+      <input className={cx({ trim })} {...props} onChange={handleChange} />
+      {!isValid && errorMessage.length && <p>{errorMessage}</p>}
+    </>
+  );
+};
+
+Input.propTypes = {
+  trim: propTypes.bool,
+};
+
 const Button = ({ label, primary, waiting, plain, ...props }) => {
   return (
     <button className={cx({ primary, waiting, plain })} {...props}>
@@ -271,14 +301,6 @@ Button.propTypes = {
   primary: propTypes.bool,
   waiting: propTypes.bool,
   plain: propTypes.bool,
-};
-
-const Input = ({ trim, ...props }) => {
-  return <input className={cx({ trim })} {...props} />;
-};
-
-Input.propTypes = {
-  trim: propTypes.bool,
 };
 
 const Tooltip = ({ children }) => {
--- a/__mocks__/@k8slens/extensions.js
+++ b/__mocks__/@k8slens/extensions.js
@@ -1,6 +1,6 @@
 import { computed, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import propTypes from 'prop-types';
 import ReactSelect, { components } from 'react-select';
 import ReactSelectCreatable from 'react-select/creatable';
@@ -259,27 +259,48 @@ class ConfirmDialog extends React.Component {
 }
 
 const Input = ({ trim, ...props }) => {
-  const [errorMessage, setErrorMessage] = useState();
-  const [isValid, setIsValid] = useState(true);
+  const [errorMessages, setErrorMessages] = useState([]);
 
-  const validate = (event) => {
-    props.validators.forEach((validator) => {
-      if (!validator.validate(event.target.value)) {
-        setIsValid(validator.validate(event.target.value));
-        setErrorMessage(validator.message());
+  const validate = (value) => {
+    const errors = [];
+
+    for (const validator of props.validators) {
+      if (errors.length) {
+        // stop validation check if there is an error already
+        break;
       }
-    });
+
+      if (!validator.validate(value)) {
+        errors.push(validator.message());
+      }
+    }
+
+    setErrorMessages(errors);
+
+    return errors;
   };
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps -- only run this on first render; run validators in handleChange() afterward
+  useEffect(() => validate(props.value), []);
+
   const handleChange = (event) => {
-    props?.onChange(event.target.value, event);
-    validate(event);
+    const isValid = !validate(event.target.value).length;
+
+    if (isValid) {
+      props?.onChange(event.target.value, event);
+    }
   };
 
   return (
     <>
       <input className={cx({ trim })} {...props} onChange={handleChange} />
-      {!isValid && errorMessage.length && <p>{errorMessage}</p>}
+      {errorMessages.length > 0 && (
+        <div className="errors">
+          {errorMessages.map((error, i) => (
+            <p key={i}>{error}</p>
+          ))}
+        </div>
+      )}
     </>
   );
 };
--- a/src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
+++ b/src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
@@ -1,13 +1,9 @@
 import mockConsole from 'jest-mock-console';
-import { render, screen, sleep, fireEvent } from 'testingUtility';
+import { render, screen } from 'testingUtility';
 import userEvent from '@testing-library/user-event';
 import { ConnectionBlock } from '../ConnectionBlock';
-import {
-  Cloud,
-  mkCloudJson,
-  CONNECTION_STATUSES,
-} from '../../../../common/Cloud'; // MOCKED
-import { CloudProvider, useClouds } from '../../../store/CloudProvider';
+import { mkCloudJson, CONNECTION_STATUSES } from '../../../../common/Cloud'; // MOCKED
+import { CloudProvider } from '../../../store/CloudProvider';
 import { IpcRenderer } from '../../../IpcRenderer';
 import { CloudStore } from '../../../../store/CloudStore';
 import * as strings from '../../../../strings';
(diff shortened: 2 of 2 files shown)


__mocks__/@k8slens/extensions.js
```diff
@@ -252,6 +252,29 @@ export class ConfirmDialog extends React.Component {
   }
 }
 
+const Input = ({ trim, onChange, ...props }) => {
+  const [inputValue, setinputValue] = useState('');
```
stefcameron, line 256: The problem with this is that the Lens component is not stateful. It doesn't internally manage the input's value as state. It requires the parent or the user of the Input to manage that state. If we make the mock stateful, then we break the rule of keep the mock as close to reality as possible.

The ConnectionBlock already has state to manage the value of the cluster name and URL inputs...

stefcameron (changes requested):
We need to figure out what's going on with that Input and `user.type()` because the way you've implemented the mock so far isn't calling the consumer's `onChange` and it's making the mock Input stateful when the real one is not.

__mocks__/@k8slens/extensions.js
AntonNos: Fixed! Thanks for helping me with that!

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,36 @@ class ConfirmDialog extends React.Component {
...
+
+  const handleChange = (event) => {
+    props?.onChange(event.target.value, event);
+    validate(event);
+  };
+
+  return (
+    <>
+      <input className={cx({ trim })} {...props} onChange={handleChange} />
+      {!isValid && errorMessage.length && <p>{errorMessage}</p>}
```
stefcameron, line 282: This should be wrapped in a `<div className="errors">`

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,36 @@ class ConfirmDialog extends React.Component {
   }
 }
 
+const Input = ({ trim, ...props }) => {
```
stefcameron, line 261: Needs an `effect(() => validate(), [])` that runs on first render only.

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,36 @@ class ConfirmDialog extends React.Component {
...
+      if (!validator.validate(event.target.value)) {
+        setIsValid(validator.validate(event.target.value));
+        setErrorMessage(validator.message());
+      }
+    });
+  };
+
+  const handleChange = (event) => {
+    props?.onChange(event.target.value, event);
+    validate(event);
```
stefcameron, line 276: Should validate first, and if there's at least one error found, do not call `onChange()`. That's how Lens does it.

AntonNos: Fixed

__mocks__/@k8slens/extensions.js
AntonNos: Fixed

__mocks__/@k8slens/extensions.js
AntonNos: Done

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,55 @@ class ConfirmDialog extends React.Component {
...
+    }
+
+    for (const validator of props.validators) {
+      if (errorMessages.length) {
+        // stop validation check if there is an error already
+        break;
+      }
+
+      if (!validator.validate(event.target.value)) {
+        setErrorMessages([errorMessages, validator.message()]);
```
stefcameron, line 276: State updates happen asynchronously, not immediately, so I'm not sure this would work all the time as expected. Also, you're missing a spread in `...errorMessages`.

We should have a local `errors = []` object in which we accumulate errors. Once the loop breaks or ends, then we `setErrorMessages(errors)` (either clearing or setting error(s)).

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,55 @@ class ConfirmDialog extends React.Component {
...
+const Input = ({ trim, ...props }) => {
+  const [errorMessages, setErrorMessages] = useState([]);
+
+  const validate = (event) => {
+    if (!event) {
+      return;
+    }
+
+    for (const validator of props.validators) {
+      if (errorMessages.length) {
```
stefcameron, line 270: Look at a local `errors` var, not the state var.

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,55 @@ class ConfirmDialog extends React.Component {
...
+        setErrorMessages([errorMessages, validator.message()]);
+      }
+    }
+  };
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => validate(), []);
+
+  const handleChange = (event) => {
+    validate(event);
```
stefcameron, line 285: Should return its local/internal `errors` list, and then we should be using that list here to determine if we'll call `onChange` or not...

src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
```diff
@@ -0,0 +1,271 @@
...
+
+      render(
+        <TestConnectionBlockComponent
+          loading={false}
+          handleClusterConnect={() => {}}
+        />
+      );
+
+      const testInvalidName = '...';
+      const inputName = document.getElementById('lecc-cluster-name');
```
stefcameron, line 171: In general, whenever I have a variable that's an element, I put `El` at the end, like `inputEl` or `inputNameEl`. When it's just `inputName`, that sounds more like it's a string than an element, which could lead someone else working in this code later on to make a wrong assumption and use it incorrectly.

src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
```diff
@@ -0,0 +1,271 @@
...
+      render(
+        <TestConnectionBlockComponent
+          loading={false}
+          handleClusterConnect={() => {}}
+        />
+      );
+
+      const testAlreadySyncedName = 'foobar';
+
+      const inputName = document.getElementById('lecc-cluster-name');
```
stefcameron, line 213: Ditto...

src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
```diff
@@ -0,0 +1,271 @@
...
+      render(
+        <TestConnectionBlockComponent
+          loading={false}
+          handleClusterConnect={() => {}}
+        />
+      );
+
+      const testAlreadySyncedUrl = 'http://barfoo.com';
+
+      const inputUrl = document.getElementById('lecc-cluster-url');
```
stefcameron, line 251: Ditto...

stefcameron (changes requested):
Nearly there!

__mocks__/@k8slens/extensions.js
AntonNos: I missed spread, thank. But anyway, we won't use that:slightly_smiling_face:
Local errors was added

__mocks__/@k8slens/extensions.js
AntonNos: Fixed

__mocks__/@k8slens/extensions.js
AntonNos: Fixed

src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
AntonNos: Will take into account, thanks! Fixed

src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
AntonNos: Fixed

src/renderer/components/GlobalPage/__tests__/ConnectionBlock.spec.js
AntonNos: Fixed

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,63 @@ class ConfirmDialog extends React.Component {
...
+      if (errors.length) {
+        // stop validation check if there is an error already
+        break;
+      }
+
+      if (!validator.validate(event.target.value)) {
+        errors.push(validator.message());
+      }
+
+      setErrorMessages(errors);
```
stefcameron, line 281: Shouldn't set this in the loop. Just once right _after_ the loop is done, just before the `return`

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,63 @@ class ConfirmDialog extends React.Component {
...
+    }
+
+    return errors;
+  };
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => validate(), []);
+
+  const handleChange = (event) => {
+    validate(event);
```
stefcameron, line 291: We're calling `validate(event)` twice now in this function... Probably this call should be dropped.

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,63 @@ class ConfirmDialog extends React.Component {
   }
 }
 
+const Input = ({ trim, ...props }) => {
+  const [errorMessages, setErrorMessages] = useState([]);
+
+  const validate = (event) => {
```
stefcameron, line 264: Instead of passing the `event` object, pass `event.target.value` or `props.value` depending on where you call `validate()` from so that validators are actually run on first render, and then anytime the value changes.

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,63 @@ class ConfirmDialog extends React.Component {
...
+    const errors = [];
+
+    if (!event) {
+      return;
+    }
```
stefcameron, line 269: Should be no need for this.

__mocks__/@k8slens/extensions.js
```diff
@@ -258,6 +258,63 @@ class ConfirmDialog extends React.Component {
...
+      }
+
+      setErrorMessages(errors);
+    }
+
+    return errors;
+  };
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => validate(), []);
```
stefcameron, line 288: Call `validate(props.value)`, which will trigger a lint error that the effect needs to depend on `[props.value]` but let's disable that error for this line only with a comment `// eslint-disable-next-line RULE_NAME -- only run this on first render; run validators in handleChange() afterward` (I forget what the rule name is...)

stefcameron (changes requested):

__mocks__/@k8slens/extensions.js
AntonNos: Right, my mistake. Fixed

__mocks__/@k8slens/extensions.js
AntonNos: Done

__mocks__/@k8slens/extensions.js
AntonNos: Done

__mocks__/@k8slens/extensions.js
AntonNos: Correct

__mocks__/@k8slens/extensions.js
AntonNos: Done

stefcameron (approved):

PR #1009, opened 2022-08-03
[PRODX-24170]: added tests to SynchronizeBlock component

AntonNos:
Screenshot 2022-08-03 at 21 16 29

### DIFF
--- a/src/renderer/components/GlobalPage/SynchronizeBlock.js
+++ b/src/renderer/components/GlobalPage/SynchronizeBlock.js
@@ -13,7 +13,6 @@ import {
   useCheckboxes,
   makeCheckboxesInitialState,
 } from '../hooks/useCheckboxes';
-import { sortNamespaces } from '../EnhancedTable/tableUtil';
 
 const { Button, Icon } = Renderer.Component;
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/GlobalPage/SynchronizeBlock.js
```diff
@@ -174,7 +173,7 @@ export const SynchronizeBlock = ({ dataCloud, onAdd }) => {
             synchronizeBlock.noProjectsFound()
           ) : (
             <ProjectsList>
-              {sortNamespaces(projectsList).map((namespace) => (
+              {projectsList.map((namespace) => (
```
stefcameron, line 176: Curious what happened here? Is it because `projectsList` is already sorted at this point?

AntonNos: We already have internal `sortByName()` method in `SynchronizeBlock` component, and with `sortNamespaces()`, sorting didn't work at all. With that `sortNamespaces()` method, even if `sortByName()` was called, projects still sorted by `asc` because of `sortNamespaces()`

AntonNos: Thats why I removed it, I think this was a bug that we didn't see

stefcameron: Got it, thanks! Good thing you wrote some tests... 😉

stefcameron (approved):
I added a CHANGELOG entry for the bug you fixed.

PR #1010, opened 2022-08-03
[PRODX-25977] Add basic structure for Create Cluster wizard

stefcameron:
Defines the steps with placeholders for content and establishes a few patterns
for how we will structure the code.

NOTE: For now, the 'Create cluster' item is only visible in dev builds.
This is so that we can easily work on the feature without making it available
if we need to publish an update, and so we don't need a feature branch.

Also relocates the cloud and namespace menu items from EnhancedTableRow to
the SyncView. EnhancedTableRow is just a component that's used by the view.
The view should be the owner of the menu items that get displayed, as well
as their handlers. In particular, the CreateClusterWizard, hence the change.

Also formally added lodash as a dependency. We were already using it, but
it wasn't declared. It must've been there as a downstream dependency of
babel-lodash.

Also updated all `mockConsole()` calls to only mock logs, infos, and
warnings, excluding errors, so that errors don't go unnoticed, since
sometimes there are errors but tests still succeed, and we shouldn't
have errors.

### Preview

cluster-wizard-start

### DIFF
--- a/__mocks__/@k8slens/extensions.js
+++ b/__mocks__/@k8slens/extensions.js
@@ -200,12 +200,12 @@ class Spinner extends React.Component {
 
 let confirmDialogInstance;
 
-export class ConfirmDialog extends React.Component {
+class ConfirmDialog extends React.Component {
   constructor(props) {
     super(props);
 
     if (confirmDialogInstance) {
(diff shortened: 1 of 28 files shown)


stefcameron:
We reviewed this together today, so I assume it's good. Merging.

PR #1015, opened 2022-08-05
[PRODX-25977] Add RTV validation to wizard

stefcameron:
This will help ensure that the data structure the wizard generates
is exactly as we expect it to be.

By using the `exactShapes` option in the verifications, we'll know
if a rogue property was added or removed somewhere in the code and
the data structure has inadvertently changed.


stefcameron:
This works well: If I add a `foobar: 1` property to the cluster step's data initializer, but don't add it to the `clusterStepTs` typeset, clicking on the Next button from the cluster step results in an RTV error exception, alerting us to the fact we added some rogue property somewhere.

AntonNos:
👀

AntonNos (approved):
Looks good!

PR #1017, opened 2022-08-08
[PRODX-24162]: added tests for catalogEntities

AntonNos:
Screenshot 2022-08-09 at 02 26 04


stefcameron:
👀

src/catalog/__tests__/catalogEntities.spec.js
```diff
@@ -0,0 +1,259 @@
+import mockConsole from 'jest-mock-console';
+import path from 'path';
+import { CatalogEntity } from '../../../__mocks__/@k8slens/extensions';
```
stefcameron, line 3: We shouldn't just get stuff from this package. This package is meant to mock the real one. So we should import `Common` from `@k8slens/extensions` (since it gets auto-mocked), and then get `CatalogEntity` like we would in real code: `const { Catalog: { CatalogEntity } } = Common;`

src/catalog/__tests__/catalogEntities.spec.js
```diff
@@ -0,0 +1,259 @@
...
+      };
+
+      it(`returns url for |${apiKind}| entity`, () => {
+        const url = `${entity.metadata.cloudUrl}/projects/${entity.metadata.namespace}`;
+
+        expect(catalogEntities.generateEntityUrl(entity)).toEqual(
+          `${url}/creds?name=${entity.metadata.name}&details`
+        );
+      });
+    });
```
stefcameron, line 74: Looks like you could combine all this into a single loop with

```
    [
      apiKinds.RHEL_LICENSE,
      apiKinds.PROXY,
      apiKinds.PUBLIC_KEY,
      apiKinds.CLUSTER,
      ...Object.values(apiCredentialKinds),
    ].forEach((apiKind) => {
```

And then use the default case to test `if (Object.values(apiCredentialKinds).includes(apiKind))` and if not, throw an error (instead of silently doing nothing) about an unexpected kind we don't have a test for yet.

src/catalog/__tests__/catalogEntities.spec.js
```diff
@@ -0,0 +1,259 @@
...
+
+      it('throw an error if entity is not instanceof CatalogEntity', () => {
+        const rhelLicenseModel = rhelLicenseObj;
+
+        expect(() =>
+          catalogEntities.entityToString(rhelLicenseModel)
+        ).toThrowError(/entity.*CatalogEntity/);
+      });
+    });
+  });
```
stefcameron, line 215: I appreciate the tests, but `modelToString()` and `entityToString()` (and the internal `_itemToString()`) are meant to be used for debug logging. We don't need to write tests for them (and writing tests would make tweaking these take more time, which wouldn't be necessary for debug/logging code).

Instead of these tests, let's ignore those 3 functions by ignoring each function, like this:
```
export const entityToString = /* istanbul ignore next -- for debug/logging only */ function (entity) {
```

stefcameron (changes requested):

src/catalog/__tests__/catalogEntities.spec.js
AntonNos: Fixed

src/catalog/__tests__/catalogEntities.spec.js
AntonNos: Makes sense, fixed

src/catalog/__tests__/catalogEntities.spec.js
AntonNos: Got it. Removed them from tests.

stefcameron:
👀

src/catalog/__tests__/catalogEntities.spec.js
```diff
@@ -0,0 +1,108 @@
...
+            break;
+          case apiKinds.CLUSTER:
+            expect(catalogEntities.generateEntityUrl(entity)).toEqual(
+              `${url}/clusters/${entity.metadata.name}`
+            );
+            break;
+          default:
+            if (!Object.values(apiCredentialKinds).includes(apiKind)) {
+              expect(catalogEntities.generateEntityUrl(entity)).toEqual(url);
+            }
```
stefcameron, line 60: Shouldn't it be this?

```suggestion
            if (Object.values(apiCredentialKinds).includes(apiKind)) {
              expect(catalogEntities.generateEntityUrl(entity)).toEqual(url);
            } else {
              throw new Error(`Unexpected apiKind="${apiKind}": Test case may need to be enhanced to support it`);
            }
```

As it is, I don't think we're testing any of the credential kinds...

src/catalog/__tests__/catalogEntities.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      const entity = {
+        metadata: {
+          cloudUrl: 'https://test-cloud-url.com',
+          kind: apiKind,
+          name: 'test-entity-name',
+          namespace: 'test-namespace',
+        },
+      };
+
+      it(`returns url for |${apiKind}| entity`, () => {
```
stefcameron, line 27: Could you add another small test case that checks that if given an entity with an unknown/unsupported kind, it just returns the basic `https://{host}/projects/{namespace}` URL? That would finish off the coverage for the `generateEntityUrl()` function.

stefcameron (changes requested):

stefcameron:
@AntonNos Should this PR be fixed and merged, or closed? It's for a ticket we have since marked as done.

AntonNos:
> @AntonNos Should this PR be fixed and merged, or closed? It's for a ticket we have since marked as done.

I fixed comments, check please

src/catalog/__tests__/catalogEntities.spec.js
```diff
@@ -0,0 +1,122 @@
...
+      const secondModel = createTestModel('uid-2');
+      const thirdModel = createTestModel('uid-3');
+
+      const listOfEntities = [firstModel, secondModel, thirdModel];
+
+      const testModel = secondModel;
+
+      expect(catalogEntities.findEntity(listOfEntities, testModel)).toEqual(
+        testModel
+      );
```
stefcameron, line 119: I think this expectation should be `toBe(testModel)` rather than `toEqual(testModel)` because `toBe` is referential equality (`===`) whereas `toEqual` will do a deep-compare of the objects, and we're expecting this function to return an item from the list, not generate a new one with the same structure.

stefcameron (changes requested):
Thanks for updating this old PR. I think every use of `toEqual()` in all tests here should actually be `toBe()`. Please try to remember to use `toBe()` _unless_ you're really deep-comparing objects/arrays, at which point use `toEqual()`. So start with `toBe()`, and if that doesn't work or doesn't fit the test case, then use `toEqual()`.

AntonNos:
> toBe

Okay, understand. Will use toBe() first in the future tests

stefcameron (approved):
Thank you! 😄

PR #1038, opened 2022-08-25
[PRODX-24168]: Added tests for topBarItems

AntonNos:
Screenshot 2022-08-26 at 00 19 27
Screenshot 2022-08-26 at 00 25 33
Screenshot 2022-08-26 at 00 25 43

OnClick handler wasn't covered, because need to update mocks.

**UPD**: OnClick handler was covered. Now we have 100% coverage
Screenshot 2022-08-26 at 12 23 22


stefcameron:
> nClick handler wasn't covered, because need to update mocks.

Actually, you don't need to update the mocks in order to test the click handler.

stefcameron:
👀

stefcameron (changes requested):
Can you think of how you can test that click handler?

AntonNos:
> Can you think of how you can test that click handler?

Yes, I already know how to do this. You are right, no need to update extension.js. My fault😉

src/renderer/__tests__/topBarItems.spec.js
```diff
@@ -0,0 +1,74 @@
...
+import userEvent from '@testing-library/user-event';
+import { ROUTE_GLOBAL_PAGE } from '../../routes';
+import * as strings from '../../strings';
+import * as topBarItems from '../topBarItems';
+
+describe('/renderer/components/EnhancedTable/TableRowListenerWrapper', () => {
+  const extension = {
+    navigate: (route) => {
+      // eslint-disable-next-line no-console --- needed for testing onClick handled in TopBarExtension
+      console.log(`Navigated to ${route}`);
```
stefcameron, line 12: This isn't how you test if a function gets call with Jest. You create a spy function instead, and you use that function as the handler. Then you check if the spy was called:

```
let extension;

beforeEach(() => {
  extension = {
    navigate: jest.fn(),
  };
});

...

it('gets called' , () => {
  render(<TopBarItem />);

  user.click('the thing');

  expect(extension.navigate).toHaveBeenCalledTimes(1);
  expect(extension.nativate.mock.calls[0][0]).toBe(expected-value-of-parameter-passed-to-function-inside-code);
});
```

src/renderer/__tests__/topBarItems.spec.js
```diff
@@ -0,0 +1,74 @@
...
+  let user;
+
+  beforeEach(() => {
+    user = userEvent.setup();
+    mockConsole(['log', 'info', 'warn']); // automatically restored after each test
+  });
+
+  describe('generateTopBarItems()', () => {
+    const TopBarItem =
+      topBarItems.generateTopBarItems(extension)[0].components.Item;
```
stefcameron, line 24: Should be in a `beforeEach()` so it's fresh for every test.

src/renderer/__tests__/topBarItems.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+      await user.click(
+        screen.getByText(strings.extension.topBar['label']()).parentNode
+      );
+
+      expect(
+        searchInMultiDim(logSpy.mock.calls, `Navigated to ${ROUTE_GLOBAL_PAGE}`)
+      ).toBe(true);
+    });
+  });
```
stefcameron, line 73: Put all this in its own `describe('First top bar item', () => ...)` block so it's obvious these tests are all related to that first top bar item in the list.

This is where you should have the `beforeEach()` that loads that first top bar item out of the list.

stefcameron (changes requested):

src/renderer/__tests__/topBarItems.spec.js
AntonNos: Thanks! Will take into account.

src/renderer/__tests__/topBarItems.spec.js
AntonNos: Done

src/renderer/__tests__/topBarItems.spec.js
AntonNos: Done

src/renderer/__tests__/topBarItems.spec.js
```diff
@@ -0,0 +1,70 @@
...
+
+      beforeEach(() => {
+        TopBarItem =
+          topBarItems.generateTopBarItems(extension)[0].components.Item;
+      });
+
+      it('renders top bar item', () => {
+        render(<TopBarItem />);
+
+        expect(topBarItems.generateTopBarItems(extension).length).toEqual(1);
```
stefcameron, line 31: This expectation should be in its own test, outside of the "First top bar item" block since it's generic to all/any top bar items that might exist. It's checking that there is exactly one top bar item that gets rendered. And then the "first top bar item" block proceeds to test that first item in the list.

stefcameron (changes requested):
Almost there, just one more tweak.

src/renderer/__tests__/topBarItems.spec.js
AntonNos: Fixed

stefcameron (approved):
Thank you!

PR #1051, opened 2022-09-02
[CCLEX-72]: added tests to ProviderStore

AntonNos:
Screenshot 2022-09-02 at 21 55 40
Screenshot 2022-09-02 at 21 55 59


stefcameron:
👀

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,77 @@
...
+import { noteOwner } from '../../../strings';
+
+describe('/renderer/store/ProviderStore', () => {
+  let pr;
+
+  const defaultStoreObj = {
+    loading: false,
+    loaded: false,
+    error: undefined,
+  };
```
stefcameron, line 12: While you're declaring a `const`, it's an object, so technically it's not constant (as in immutable). If it was just a primitive like string, number, boolean, null, undefined, then it would be fine. But when it's an object (or an array), it's always safer to redefine it in a `beforeEach()` block so it's new for every test, just in case a test inadvertently/unexpectedly ends-up modifying it for some reason.

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    pr.loaded = true;
+    expect(pr.loaded).toEqual(true);
+  });
+
+  it('generates a new, empty store object', () => {
+    expect(pr.makeNew()).toMatchObject(defaultStoreObj);
+  });
+
+  it('creates a SHALLOW clone of store object', () => {
+    expect(pr.clone()).toMatchObject(defaultStoreObj);
```
stefcameron, line 35: This would be a stronger test if the store contained either an array or an object property, and then you checked that this property, in the clone, is equal to the property in the original using referential equality (`===`).

Note that for this, you'll need to make a test that creates a class that inherits from `ProviderStore` and overrides `makeNew()` in order to add a new property to the ProviderStore's `store` property. Don't forget to call `super.makeNew()`. See https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer/store/CloudProvider.js#L14 -- incidentally, looks like we have a bug in CloudProvider in that the `makeNew()` override is _not_ calling `super.makeNew()` like it should.

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,77 @@
...
+
+  it('posts an error notification to the UI with its content', () => {
+    expect(document.querySelector('.message')).not.toBeInTheDocument();
+
+    pr.error = 'some error message';
+    pr.notifyIfError();
+
+    expect(document.querySelector('.message').innerHTML).toEqual(
+      `${pr.error} ${noteOwner}`
+    );
```
stefcameron, line 75:
```suggestion
    expect(document.querySelector('.message')).toHaveTextContent(
      `${pr.error} ${noteOwner}`
    );
```

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+  });
+
+  it('resets store state, loading=|false|', () => {
+    pr.setState = () => {};
+
+    expect(pr.store).toMatchObject(defaultStoreObj);
+    pr.reset();
+    expect(pr.store).toMatchObject(defaultStoreObj);
+  });
```
stefcameron, line 59: Same as the previous test: These would be stronger tests if you had a class that inherits from `ProviderStore` and you could also validate that the extended class' `makeNew()` was called as a result of calling `pr.reset()` in order to generate new values for the store properties. No need to mock the extended class' `makeNew()`. Just by creating the store, then altering the property that the extended class sets on the store to something other than its default value, and then calling `pr.reset()` should result in the store's value for that property to be back to its original value...

stefcameron, line 59: Same as the previous test: These would be stronger tests if you had a class that inherits from `ProviderStore` and you could also validate that the extended class' `makeNew()` was called as a result of calling `pr.reset()` in order to generate new values for the store properties. No need to mock the extended class' `makeNew()`. Just by creating the store, then altering the property that the extended class sets on the store to something other than its default value, and then calling `pr.reset()` should result in the store's value for that property to be back to its original value...

stefcameron (changes requested):
71.4% is too low... We want all 4 categories to be above 85%.

If that `|| undefined` branch is the only thing left, you should be able to write a simple test for that like
```javascript
pr.error = '';
expect(pr.store.error).toBeUndefined();
```

AntonNos:
> 71.4% is too low... We want all 4 categories to be above 85%.
> 
> If that `|| undefined` branch is the only thing left, you should be able to write a simple test for that like
> 
> ```js
> pr.error = '';
> expect(pr.store.error).toBeUndefined();
> ```

All comments are fixed.

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,149 @@
...
+
+  describe('clone()', () => {
+    class TestProviderStore extends ProviderStore {
+      constructor(props) {
+        super(props);
+      }
+
+      makeNew() {
+        return {
```
stefcameron, line 69: Should be calling `super.makeNew()` here -- also need to fix this in CloudProvider.js since it it not -- and the object you return should add the `testProperty` to the object you got from super...

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,149 @@
...
+      storeObj = {
+        loading: false,
+        loaded: false,
+        error: undefined,
+        testProperty: [],
+      };
+    });
+
+    it('creates a SHALLOW clone of store object', () => {
+      expect(tps.clone().testProperty).toEqual(storeObj.testProperty);
```
stefcameron, line 93: `toEqual()` does not compare using referential equality (`===`) -- that is, checking that the left and right sides are equal, not just in the same structure (i.e. two arrays with the same elements), but also as being the same objects in _memory_. You should use `toBe()` here. `toEqual()` is like `==` whereas `toBe()` is like `===`.

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,149 @@
...
+    class TestProviderStore extends ProviderStore {
+      constructor(props) {
+        super(props);
+      }
+
+      makeNew() {
+        return {
+          loading: false,
+          loaded: false,
+          error: undefined,
```
stefcameron, line 107: These are provided by the base class. No need to have `makeNew()` at all in this case since you aren't adding any custom properties.

But to address my previous comment about these tests here, you should keep `makeNew()`, call the superclass, and then add an additional custom property -- like what you're doing in your `clone()` tests.

Then you should set that property to some new value (`pr.store.testProperty = [1, 2, 3]`), reset the store, and then validate that `pr.store.testProperty` is back to its original value of an empty array...

stefcameron (changes requested):

src/renderer/store/__tests__/ProviderStore.spec.js
AntonNos: > Should be calling `super.makeNew()` here -- also need to fix this in CloudProvider.js since it it not -- and the object you return should add the `testProperty` to the object you got from super...

All comments are fixed

AntonNos: Fixed

src/renderer/store/__tests__/ProviderStore.spec.js
AntonNos: Got it, fixed

AntonNos:
@stefcameron Comments are fixed, check please.

stefcameron:
👀

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    beforeEach(() => {
+      tps = new TestProviderStore();
+
+      storeObj = {
+        ...tps.clone(),
+      };
+    });
+
+    it('creates a SHALLOW clone of store object', () => {
+      expect(tps.clone().testProperty).toBe(storeObj.testProperty);
```
stefcameron, line 88: Wouldn't this work, and be simpler?
```suggestion
      expect(tps.clone().testProperty).toBe(tps.store.testProperty);
```

src/renderer/store/__tests__/ProviderStore.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    it('resets store state, loading=|true|', () => {
+      const newLoading = true;
+
+      tps.setState = () => {};
+      tps.store.testProperty = [1, 2, 3];
+
+      tps.reset(newLoading);
+
+      expect(spy).toHaveBeenCalled();
+      expect(tps.store.testProperty.length).toEqual(0);
```
stefcameron, line 123: The proper assertion here would be
```suggestion
      expect(tps.store.testProperty).toHaveLength(0);
```

stefcameron (approved):
Thanks for fixing that issue in CloudProvider too.

I pointed out 2 last things, but I don't need to review again if you decide to make those changes. Provided it builds, you can merge it.

AntonNos:
> Thanks for fixing that issue in CloudProvider too.
> 
> I pointed out 2 last things, but I don't need to review again if you decide to make those changes. Provided it builds, you can merge it.

Thanks! I fixed them and going to merge.

PR #1092, opened 2022-09-21
[CCLEX-165]: added last sync time stamp to entity details view

AntonNos:
Screenshot 2022-09-22 at 02 13 52


stefcameron:
👀

src/api/types/Resource.js
```diff
@@ -48,6 +48,8 @@ export class Resource {
         }
       );
 
+    const syncedDate = new Date();
```
stefcameron, line 51: Since this is meant to be a _private_ member, could you rename it to `_syncedDate` to follow the pattern in other classes like SyncManager, DataCloud, Cloud, etc?

src/renderer/catalogEntityDetails.js
```diff
@@ -14,10 +14,10 @@ import { openBrowser } from '../util/netUtil';
 
 dayjs.extend(dayjsRelativeTimePlugin);
 
-const getCreatedValue = (createdAt) => {
-  return `${dayjs(createdAt).format('MMMM DD, YYYY')} (${dayjs(
-    createdAt
-  ).fromNow(true)})`;
+const getDataValue = (date) => {
```
stefcameron, line 17: Did you meant to name it `getDateValue()`?

src/strings.ts
```diff
@@ -222,8 +222,10 @@ export const catalog: Dict = {
         unknownValue: () => '<Unknown>', // when a value is null/undefined/empty
         props: {
           dateCreated: () => 'Date created',
+          lastSync: () => 'Last sync',
           uid: () => 'UID',
           serverStatus: () => `${mccCodeName} status`,
+          le: () => 'Lens Extension',
```
stefcameron, line 228: `le` is convenient, but not meaningful when reading the code elsewhere. 🙂 Maybe `lensExtension`?

stefcameron (changes requested):
Looks great overall, nice work! Just some minor naming requests.

src/renderer/catalogEntityDetails.js
```diff
@@ -14,10 +14,10 @@ import { openBrowser } from '../util/netUtil';
...
+  return `${dayjs(date).format('MMMM DD, YYYY')} (${dayjs(date).fromNow(
```
stefcameron, line 18: I think we should add the hours, minutes, seconds and reduce the full month name to the abbreviation (I think that will be `MMM` instead of `MMMM`). Right now, we get "September 22, 2022 (a few minutes ago)` which isn't precise enough.

src/catalog/catalogEntities.js
```diff
@@ -81,6 +81,7 @@ export const catalogEntityModelTs = {
     //  if necessary
     namespace: rtv.STRING,
     cloudUrl: [rtv.STRING, (v) => !v.endsWith('/')], // no trailing slash
+    syncedAt: rtv.STRING,
```
stefcameron, line 84: Adding this is good, but I see now that it's causing a change to the schema, which is causing an exception in the SyncStore.js [right here](https://github.com/Mirantis/lens-extension-cc/blob/master/src/store/SyncStore.js#L112) when it loads existing data from sync-store.json when Lens starts, which is resulting in the loss of all previously-synced items in the catalog if the DataCloud is disconnected/offline on load.

Basically, if a customer installs a newer version of this extension with this change in it, and their clouds are disconnected or they're offline somewhere, they won't see any of their items in their catalog.

So we need to "upgrade" those entities in the sync-store.json on load before validating them to avoid this issue.

One way would be to make this property optional (i.e. `syncedAt: [rtv.OPTIONAL, rtv.STRING]`) but the side effects of that is that we have to code defensively everywhere we want to use this property, and in reality, it's not really an optional property.

We may have other cases like this in the future, so how do you think we could handle this? Let's discuss before you implement a solution.

stefcameron (changes requested):

src/api/types/Resource.js
AntonNos: Okay, will do that also in future tasks.

src/renderer/catalogEntityDetails.js
AntonNos: Yes, thank you! My fault with typo

src/strings.ts
AntonNos: Fixed

src/renderer/catalogEntityDetails.js
AntonNos: Added

src/catalog/catalogEntities.js
AntonNos: Hmm, I thought, that it should be synced at least once, when we connect/sync cloud in Lens. So I thought, that we will always have `syncedAt` as a string with date.

src/renderer/catalogEntityDetails.js
```diff
@@ -14,9 +14,9 @@ import { openBrowser } from '../util/netUtil';
 
 dayjs.extend(dayjsRelativeTimePlugin);
 
-const getCreatedValue = (createdAt) => {
-  return `${dayjs(createdAt).format('MMMM DD, YYYY')} (${dayjs(
-    createdAt
+const getDateValue = (date) => {
+  return `${dayjs(date).format('MMM DD, YYYY, HH:mm:ss')} (${dayjs(
```
stefcameron, line 18: Is that 24-hour or 12-hour time? Should be 24-hour.

src/catalog/catalogEntities.js
stefcameron: Not for an existing customer upgrading to a newer version of the extension. Do you think there's something we could do in `SyncStore.js`...?

Keep in mind, this type of enhancement likely won't be the last we'll make, and we may get customers upgrading from varying older versions of the extension (like `6 -> 6.1` or `6.3 -> 6.10`, for example).

src/renderer/catalogEntityDetails.js
AntonNos: Screenshot 2022-09-23 at 12 54 33
 It's 24-hour time.

src/catalog/catalogEntities.js
AntonNos: Maybe in `getDefaults()`, we can add `syncedAt` property to each entity with default value (empty string??)?

stefcameron: That's along the lines of what I was thinking as well. But putting it in `getDefaults()` wouldn't work. How would you know all the entities to create, for each individual customer's installation...?

So I think we should add a function, `upgradeEntities(store)`, that we call at [line 110](https://github.com/Mirantis/lens-extension-cc/blob/master/src/store/SyncStore.js#L110). It's job will be to __cautiously__ iterate (i.e. use tests like optional chaining and IF statements to prevent crashes) through the store and look for `syncedAt` properties. If it's undefined, it will add it.

This way, when line 112 executes, the schema will be valid, and we can rely on that property always existing.

And I think if we're adding the property, since we don't know when it was last synced, we should create it as `new Date(0)` (i.e. the 1970/1/1 epoch when the Date object begins) so it's obvious it's old and may be outdated.

AntonNos: Thank you, let me check it.

AntonNos: Added new upgradeEntities() method

src/store/SyncStore.js
```diff
@@ -102,11 +102,26 @@ export class SyncStore extends Common.Store.ExtensionStore {
     );
   }
 
+  /**
+   * Add syncedAt property to entity metadata if it's undefined
+   */
+  upgradeEntities(store) {
+    return Object.values(store).map((entity) => {
```
stefcameron, line 109: No need to return the resulting array, right?

src/store/SyncStore.js
```diff
@@ -102,11 +102,26 @@ export class SyncStore extends Common.Store.ExtensionStore {
...
+      entity.map((singleEntity) => {
+        if (!singleEntity.metadata.syncedAt) {
+          singleEntity.metadata.syncedAt = new Date(0).toISOString();
+        }
+      });
```
stefcameron, line 114: Since we're pre-validation here, we need to be more careful so we don't just crash on startup if the shape of the store isn't exactly as we expect it to be. Also, the store is expected to be a map of entity type to __array__ of entities of that type... And for each entity, we don't know if it'll be defined, and if it'll have a `metadata` property that's an object (remember, `typeof null === 'object'` in JS) on which we can check for a property and set a property since we can't do that on a primitive.

Our goal here should be to try our best to add that `syncedAt` property, if it's not defined, such that the call to `rtv.check({ store }, ...)` will succeed.

stefcameron (changes requested):

src/store/SyncStore.js
AntonNos: Yes, you are right, thanks.

src/store/SyncStore.js
AntonNos: Thanks for comment, I fixed it.

src/renderer/catalogEntityDetails.js
stefcameron, line 18: I think one more thing that would make this nicer is to check for `date.getTime() === 0` and use the string "Unknown" instead of generating the related date that's like 52.5 years ago now.

AntonNos: Makes sense, fixed.

src/renderer/catalogEntityDetails.js
```diff
@@ -14,10 +14,12 @@ import { openBrowser } from '../util/netUtil';
...
-const getCreatedValue = (createdAt) => {
-  return `${dayjs(createdAt).format('MMMM DD, YYYY')} (${dayjs(
-    createdAt
-  ).fromNow(true)})`;
+const getDateValue = (date) => {
+  return date.getTime() === 0
+    ? `${dayjs(date).format('MMM DD, YYYY, HH:mm:ss')} (${dayjs(date).fromNow(
+        true
+      )})`
+    : 'Unknown';
```
stefcameron, line 22: Please use a string from `strings.js`... 🙂  I think we can use our existing one: https://github.com/Mirantis/lens-extension-cc/blob/master/src/strings.ts#L222

src/renderer/catalogEntityDetails.js
```diff
@@ -14,10 +14,12 @@ import { openBrowser } from '../util/netUtil';
 
 dayjs.extend(dayjsRelativeTimePlugin);
 
-const getCreatedValue = (createdAt) => {
-  return `${dayjs(createdAt).format('MMMM DD, YYYY')} (${dayjs(
-    createdAt
-  ).fromNow(true)})`;
+const getDateValue = (date) => {
+  return date.getTime() === 0
+    ? `${dayjs(date).format('MMM DD, YYYY, HH:mm:ss')} (${dayjs(date).fromNow(
```
stefcameron, line 19: The logic is reversed: It will show "Unknown" for all cases _except_ when it's the JS epoch... Make sure you test it at least manually...

stefcameron (changes requested):

src/renderer/catalogEntityDetails.js
AntonNos: Sorry for such mistakes, I'm already falling asleep at the computer. Going to fix them right now

src/renderer/catalogEntityDetails.js
AntonNos: Right, thank you

src/renderer/catalogEntityDetails.js
```diff
@@ -34,6 +24,18 @@ const {
   },
 } = strings;
 
+const getDateValue = (date) => {
+  return date.getTime() === 0
```
stefcameron, line 28: This is better, but if you run this, it crashes... because what you're passing to `getDateValue()` is a __string__ (either `spec.createdAt` or `metadata.syncedAt`), not a `Date` object. The `daysjs()` function handles a string and converts it to a `Date` for you. I didn't see it myself, until I ran the code. So probably this function would be better/safer as:
```javascript
const getDateValue = (date) => {
  const dateObj = new Date(date);
  return dateObj.getTime() === 0 ? unknownValue() : `...`;
};
```

stefcameron (changes requested):
Go to bed! 🙂 You can finish this once you get some rest.

src/renderer/catalogEntityDetails.js
AntonNos: You are right, thanks for solution! Fixed.

stefcameron (approved):
👍

PR #1120, opened 2022-10-07
[CCLEX-171] Skip TLS verification in kubeConfigs when requested

stefcameron:
When setting `LEX_CC_UNSAFE_ALLOW_SELFSIGNED_CERTS=1` on the command line to run Lens, we disable TLS verification when making API requests, but we aren't disabling it when Lens attempts to access the cluster to generate a cluster token via the `kube-login` binary.

This fix will disable TLS verification in the kubeConfig also, if the command line flag is set.

### PR Checklist

Check if done, remove if not applicable:


AntonNos (approved):
👍

PR #1135, opened 2022-10-17
[CCLEX-65]: added tests for apiUtil

AntonNos:
Screenshot 2022-10-17 at 13 22 56
Screenshot 2022-10-17 at 13 23 25

**Update:**
Covered 85+% in each block.
Screenshot 2022-10-31 at 23 52 46

### DIFF
--- a/src/api/__tests__/apiUtil.spec.js
+++ b/src/api/__tests__/apiUtil.spec.js
@@ -78,7 +78,7 @@ describe('/api/apiUtil', () => {
       refreshToken: REFRESH_TOKENS.VALID,
     });
 
-    mockConsole(['log', 'info', 'warn', 'error']); // automatically restored after each test
+    mockConsole(['log', 'info', 'warn']); // automatically restored after each test
   });
 
   describe('extractJwtPayload()', () => {
--- a/src/api/__tests__/apiUtil.spec.js
+++ b/src/api/__tests__/apiUtil.spec.js
@@ -104,11 +104,16 @@ describe('/api/apiUtil', () => {
     });
 
     describe('errors', () => {
+      const logSpy = jest.spyOn(console, 'error');
+
       it('triggers an error', async () => {
         isTokensRefreshed = await apiUtil.cloudRefresh(
           fakeCloudWithInvalidRefreshToken
         );
         expect(isTokensRefreshed).toBe(false);
+        expect(logSpy.mock.calls[0][0]).toMatch(
+          /(Unable to refresh expired token)/i
+        );
       });
     });
 
--- a/src/api/__tests__/apiUtil.spec.js
+++ b/src/api/__tests__/apiUtil.spec.js
@@ -10,6 +10,7 @@ const REFRESH_TOKENS = {
 };
 
 const testErrorMessage = 'error-message';
+const testSuccessBodyMessage = 'success';
 
 jest.mock('../../common/Cloud');
 jest.mock('../clients/ApiClient', () => {
@@ -18,7 +19,9 @@ jest.mock('../clients/ApiClient', () => {
   const ApiClient = class {
     refreshToken(refreshToken) {
       return Promise.resolve({
-        responce: {},
+        response: {
+          status: refreshToken === REFRESH_TOKENS.VALID ? 200 : 400,
+        },
         body: '',
         error: refreshToken === REFRESH_TOKENS.VALID ? '' : testErrorMessage,
       });
@@ -34,6 +37,28 @@ jest.mock('../clients/ApiClient', () => {
   return {
     ...original,
     ApiClient,
+  };
+});
+jest.mock('../clients/ResourceClient', () => {
+  const original = jest.requireActual('../clients/ResourceClient');
+
+  const ResourceClient = class {
+    list(resourceType) {
+      return Promise.resolve({
+        response: {
+          status: resourceType === 'vspherecredentials' ? 401 : 200,
+        },
+        error: resourceType === 'byocredentials' ? testErrorMessage : '',
+        body:
+          resourceType === 'vspherecredentials' ? '' : testSuccessBodyMessage,
+        url: '',
+      });
+    }
+  };
+
+  return {
+    ...original,
+    ResourceClient,
   };
 });
 
@@ -53,7 +78,7 @@ describe('/api/apiUtil', () => {
       refreshToken: REFRESH_TOKENS.VALID,
     });
 
-    // mockConsole(['log', 'info', 'warn', 'error']); // automatically restored after each test
+    mockConsole(['log', 'info', 'warn', 'error']); // automatically restored after each test
   });
 
   describe('extractJwtPayload()', () => {
@@ -66,8 +91,8 @@ describe('/api/apiUtil', () => {
 
     it('returns a JSON object if token provided', () => {
       expect(apiUtil.extractJwtPayload(testToken)).toMatchObject({
-        'foo': 123
-      })
+        foo: 123,
+      });
     });
   });
 
@@ -126,7 +151,6 @@ describe('/api/apiUtil', () => {
 
   describe('cloudRequest()', () => {
     let fakeCloudWithoutToken;
-    let fakeCloudWithToken;
     let fakeCloud;
 
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


stefcameron:
👀

stefcameron:
Thanks for your on these tests! What we're aiming for is 85% on all 4 aspects here:
image

Statements and Lines are still below 85%. Can you write a test to go into the IF block at line 177?

src/api/__tests__/apiUtil.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      cloudUrl: cloudUrl,
+      refreshToken: REFRESH_TOKENS.INVALID,
+    });
+
+    fakeCloudWithValidRefreshToken = new Cloud({
+      cloudUrl: cloudUrl,
+      refreshToken: REFRESH_TOKENS.VALID,
+    });
+
+    mockConsole(['log', 'info', 'warn', 'error']); // automatically restored after each test
```
stefcameron, line 56: I think in most other test modules, we don't mock `error` so that if there are errors, they aren't prevented from surfacing in the logs -- unless we're specifically testing that an error message is logged using `console.error()`, of course.

src/api/__tests__/apiUtil.spec.js
```diff
@@ -0,0 +1,180 @@
...
...0MTkwMDM2LCJqdGkiOiIwYmJhYzQ3Yi1hN2Q5LTQyZTEtYWYzMy05OTdhNTA5NTg3OTgiLCJpc3MiOiJodHRwczovL2NvbnRhaW5lci1jbG91ZC1hdXRoLmludC5taXJhbnRpcy5jb20vYXV0aC9yZWFsbXMvaWFtIiwiYXVkIjoia2FhcyIsInN1YiI6ImU0OTU2MmE5LWMzMDAtNDRlMy1hYTJiLTgwYzkyZmQ2YmYyMyIsInR5cCI6IklEIiwiYXpwIjoia2FhcyIsInNlc3Npb25fc3RhdGUiOiIxN2QyMGVkMy1mODYxLTQ4MDMtOTE5Mi0zYjJmZjllZDNkMzYiLCJhdF9oYXNoIjoiNVNXZlZON0EyaEZjUXhRMUtNOTNKUSIsInNpZCI6IjE3ZDIwZWQzLWY4NjEtNDgwMy05MTkyLTNiMmZmOWVkM2QzNiIsImlhbV9yb2xlcyI6WyJtOmthYXM6aW1jLW1jYy11eC10ZWFtQHdyaXRlciIsIm06a2FhczpsZXgtbnMtMUB3cml0ZXIiLCJtOmthYXM6aW1jLWs4cy10ZWFtQHdyaXRlciIsIm06a2FhczpsZXgtbnMtMkB3cml0ZXIiLCJtOmthYXM6ZGVmYXVsdEByZWFkZXIiXSwiZW1haWxfdmVyaWZpZWQiOmZhbHNlLCJuYW1lIjoiQW50b24gTm9zIiwicHJlZmVycmVkX3VzZXJuYW1lIjoiYW5vc0BtaXJhbnRpcy5jb20iLCJnaXZlbl9uYW1lIjoiQW50b24iLCJmYW1pbHlfbmFtZSI6Ik5vcyIsImVtYWlsIjoiYW5vc0BtaXJhbnRpcy5jb20ifQ.Vh7mHeZ8LpvLZfXoGcDPq9g1xAc3SQpfwLck4k__KurdYIxkpE2xSEQBwnJl-VhNmXvssfYsOpObNS_-upS32GSus-q6rHvl7t4hV6nofpwBVmAWNbuLmiqD667y2zyjfhow4_3QSI59ZfOM7A0AHpQJLf6sFAcGyf_nYSHVjZueVh9wvBkD1vK7fe9JggoCXmT3GBW0wf5fhqM862-BJVMWV3Y1nhURtPnowdqIRfR9_EOLVHxE1ewAWteJGTIsLO7nV59dWUIGFfd_C3hKOAoCW8EqRqOmYuJ-aiNZxR-IUgIXqgUZw250aYZX8F0kVaKT7xtczGj1nvoUlazJxg';
```
stefcameron, line 61: This must be a real JWT. Where did it come from? Just want to make sure you're not using an actual token for security reasons... You can use https://jwt.io/ to decode it by pasting it on the left hand side. And you can use the right hand side to edit it or create one with any content you want. So you should generate one with a payload like `{ foo: 123 }` and test at line 69 that you get an object with the expected property and value back from the method instead of just testing you get an object with some properties.

src/api/__tests__/apiUtil.spec.js
```diff
@@ -0,0 +1,180 @@
...
+        const result = await apiUtil.cloudRequest({
+          cloud: fakeCloudWithoutToken,
+          method: 'list',
+          resourceType: apiResourceTypes.NAMESPACE,
+        });
+
+        expect(result.error).toBe(strings.apiUtil.error.noTokens());
+      });
+
+      it('triggers an error when invalid resourse type', async () => {
```
stefcameron, line 153: Typo: "resourse" -> "resource"

src/api/__tests__/apiUtil.spec.js
```diff
@@ -0,0 +1,180 @@
...
+
+    describe('successes', () => {
+      it('makes an authenticated request using the given Cloud', async () => {
+        const result = await apiUtil.cloudRequest({
+          cloud: fakeCloud,
+          method: 'list',
+          resourceType: apiResourceTypes.NAMESPACE,
+        });
+
+        expect(Object.keys(result)).toHaveLength(7);
```
stefcameron, line 176: In this case, we expect the request to succeed, so it would be a much stronger test to assert that the given payload looks like a success rather than error. The way you've done it so far, something could break causing an error to be returned and the test might continue to think that everything's working.

stefcameron (changes requested):

src/api/__tests__/apiUtil.spec.js
AntonNos: Screenshot 2022-10-25 at 12 34 11
I mocked `error` here, because in one of the test, I expected to have that error in console, and it thrown using `logger.error`

src/api/__tests__/apiUtil.spec.js
AntonNos: Thanks for letting me know about https://jwt.io/! Will use it in future work (if need).
I fixed test.

src/api/__tests__/apiUtil.spec.js
AntonNos: Fixed

stefcameron:
👀

stefcameron:
@AntonNos I went looking for updates based on your comments, but I don't see any. Did you forget to push your commit?

AntonNos:
> @AntonNos I went looking for updates based on your comments, but I don't see any. Did you forget to push your commit?

Yes, I didn't push it, because not all comments was fixed. I had changes on my PC and once all comments will be fixed - I will push update and will re-request review.

src/api/__tests__/apiUtil.spec.js
stefcameron: The tests are looking great now, and I think you removed the expectation related to the error log, right? I don't see an `expect(console.error).toHaveBeenCalled()` anywhere (which is good). So can we remove 'error' from the list here like in the other test specs?

stefcameron (commented):
Looks good, just wondering about that `mockConsole()` configuration (see earlier comment thread).

src/api/__tests__/apiUtil.spec.js
AntonNos: I checked it 1 more time, and without mocking 'error', we received it in console (logger.error is in the code). So I added 1 more expectation to tests. Check please

AntonNos: Hmm, with mocked error test with error expectation fails. I removed it from mockConsole.

stefcameron (approved):

src/api/__tests__/apiUtil.spec.js
stefcameron: I don't see any test that has `expect(console.error)...` so I'm not sure what you mean. I see you removed "error" from `mockConsole()`, so AFAICT, we're good. 🙂

PR #1152, opened 2022-10-24
[CCLEX-172]: added Overview/Summary panel to Cluster page

AntonNos:
**Management cluster Overview panel:**
Screenshot 2022-10-27 at 14 40 53

**Child cluster Overview panel:**
Screenshot 2022-10-27 at 14 40 16

**Light mode:**
Screenshot 2022-11-01 at 16 11 40

**FYI:** Tests for `ClusterOverviewPage` and `SummaryPanel` components wasn't written, because need to modify `extension.js` (we are using `clusterEntity` from `Renderer.Catalog.catalogEntities` in `ClusterOverviewPage` and `DrawerTitle`/`DrawerItem` components in `SummaryPanel`).


stefcameron:
👀

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,373 @@
...
+        return (
+          <ProviderWrapper>
+            <IconWrapper>
+              <VsphereIcon size={8} fill="var(--textColorPrimary)" />
+            </IconWrapper>
+            {provider}
+          </ProviderWrapper>
+        );
+      default:
+        return <>{strings.catalog.entities.common.details.unknownValue()}</>;
```
stefcameron, line 146: I think in this module, we've already imported as just `unknownValue()`

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,373 @@
...
+                      getClusterLabelsCount(
+                        clusterEntity.metadata.labels['credential']
+                      )
+                    )}
+                    :
+                  </EntityCell>
+                  <td>{clusterEntity.metadata.labels['credential'] || '--'}</td>
+                </tr>
+              </tbody>
+            </table>
```
stefcameron, line 253: Tables are generally awkward. Could you just use a flex container instead? Same for each other item below...

stefcameron (changes requested):
Screenshots look good! In general, let's avoid using tables whenever possible. Flex containers are a great lightweight alternative and I think easier to work with and style.

- [x] Can you add screenshots of what this looks like in Light mode?

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,373 @@
...
+   */
+  const getProvider = (provider) => {
+    switch (provider) {
+      case consts.providerTypes.AWS:
+        return (
+          <ProviderWrapper>
+            <IconWrapper>
+              <AwsIcon size={17} fill="var(--textColorPrimary)" />
+            </IconWrapper>
+            {provider}
```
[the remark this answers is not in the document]
AntonNos: Removed. It was added, because I saw that in Figma we had a string near icon.

src/renderer/catalogEntityDetails.js
```diff
@@ -24,7 +24,7 @@ const {
   },
 } = strings;
 
-const getDateValue = (date) => {
+export const getDateValue = (date) => {
```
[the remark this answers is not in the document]
AntonNos: Makes sense. Done

src/strings.ts
```diff
@@ -206,6 +206,26 @@ export const clusterPage: Dict = {
...
+        releaseVersion: () => 'Release version',
+        namespace: () => 'Namespace',
+        clusterUrl: () => 'Cluster URL',
+        clusterObjects: {
+          title: () => 'Cluster objects',
+          credentials: (count) => (count > 1 ? 'Credentials' : 'Credential'),
+          sshKeys: (count) => (count > 1 ? 'SSH keys' : 'SSH key'),
+          rhelLicenses: (count) =>
+            count > 1 ? 'RHEL licenses' : 'RHEL license',
+          proxies: (count) => (count > 1 ? 'Proxies' : 'Proxy'),
```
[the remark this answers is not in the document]
AntonNos: Got it. I didn't know that 0 is plural😄

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,373 @@
...
+  //
+  // EVENTS
+  //
+
+  /**
+   * Returns provider icon with it name.
+   * @param {string} provider Provider name.
+   * @returns {HTMLElement} Provider info, including icon and name.
+   */
+  const getProvider = (provider) => {
```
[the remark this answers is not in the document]
AntonNos: You are right! Moved it outside

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,373 @@
...
+          {clusterEntity.metadata.syncedAt
+            ? getDateValue(clusterEntity.metadata.syncedAt)
+            : unknownValue()}
+        </DrawerItem>
+        <DrawerItem
+          name={strings.clusterPage.pages.overview.summary.provider()}
+        >
+          {clusterEntity.spec.provider
+            ? getProvider(clusterEntity.spec.provider)
+            : unknownValue()}
```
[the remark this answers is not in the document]
AntonNos: Yes, missed that

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Thanks, fixed.

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Yes, sure! I forgot to get rid of them after changing tables to DrawerItem

src/strings.ts
```diff
@@ -206,6 +206,26 @@ export const clusterPage: Dict = {
...
+        clusterObjects: {
+          title: () => 'Cluster objects',
+          credentials: (count) => (count === 1 ? 'Credential' : 'Credentials'),
+          sshKeys: (count) => (count === 1 ? 'SSH key' : 'SSH keys'),
+          rhelLicenses: (count) =>
+            count === 1 ? 'RHEL license' : 'RHEL licenses',
+          proxies: (count) => (count === 1 ? 'Proxy' : 'Proxies'),
+        },
+      },
     },
```
stefcameron, line 229: This is mostly (not all, but most) a duplicate of the labels for the `Catalog > Cluster > View details` panel strings here: https://github.com/Mirantis/lens-extension-cc/blob/master/src/strings.ts#L259-L282

Let's move lines 197 to 231 down past the existing `export const catalog: Dict = {` definition, and then refer to those strings like this:

```javascript
      summary: {
        title: () => 'Summary',
        clusterName: () => 'Cluster name',
        serverStatus: () => catalog.entities.common.details.props.serverStatus,
        syncTime: () => catalog.entities.common.details.props.lastSync,
        provider: () => catalog.entities.cluster.details.props.provider,
        managementCluster: catalog.entities.cluster.details.props.managementCluster,
        isManagementCluster: catalog.entities.cluster.details.props.isManagementCluster,
        releaseVersion: () => catalog.entities.cluster.details.props.release,
        namespace: () => 'Namespace',
        clusterUrl: () => 'Cluster URL',
        clusterObjects: {
          title: () => 'Cluster objects',
          credentials: (count) => (count === 1 ? 'Credential' : 'Credentials'),
          sshKeys: (count) => (count === 1 ? 'SSH key' : 'SSH keys'),
          rhelLicenses: (count) =>
            count === 1 ? 'RHEL license' : 'RHEL licenses',
          proxies: (count) => (count === 1 ? 'Proxy' : 'Proxies'),
        },
      },
```

You'll want to fill in the rest in your Details PR...

Also, instead of `mccStatus`, let's use `serverStatus` (trying to avoid using a marketing term in the name of a variable/property) (I renamed it above already).

Also, `syncTime` should be `lastSync` with label "Last sync" like we have in the View Details panel -- we should keep the same labels when possible. I fixed/updated it in the mock since it was "Sync time" there instead of "Last sync".

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,343 @@
...
+  marginRight: layout.grid * 3.5,
+  minWidth: layout.grid * 32.5,
+}));
+
+const ProviderWrapper = styled.span(() => ({
+  display: 'flex',
+  alignItems: 'center',
+}));
+
+const MccStatus = styled.p`
```
stefcameron, line 73: Can we name this `ServerStatus` instead? I'd like to avoid using marketing terms like `mcc` in variable names as much as possible.

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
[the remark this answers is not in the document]
stefcameron: Still needs fixing as far as I can see in the code.

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,343 @@
...
+                  <Icon material="verified_user" />
+                </IconWrapper>
+                {strings.clusterPage.pages.overview.summary.clusterObjects.credentials(
+                  getClusterLabelsCount(
+                    clusterEntity.metadata.labels['credential']
+                  )
+                )}
+                :
+              </EntityWrapper>
+              <span>{clusterEntity.metadata.labels['credential'] || '--'}</span>
```
stefcameron, line 236: Instead of hard coding the empty `"--"` string value, let's put it in the strings in `catalog.entities.common.details.emptyValue: () => '--'` like we have for `unknownValue()` so we can re-use it consistently across this, the Cluster Page Details panel, and also the Cluster View Details panel...

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,343 @@
...
+}));
+
+const EntityWrapper = styled.span(() => ({
+  display: 'inline-flex',
+  alignItems: 'center',
+  marginRight: layout.grid * 3.5,
+  minWidth: layout.grid * 32.5,
+}));
+
+const ProviderWrapper = styled.span(() => ({
```
stefcameron, line 68: Looks like you're using a `<span>` as a container, which isn't legal HTML. Seems like this should be a `<div>`.

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,343 @@
...
+const ClusterLink = styled.a(() => ({
+  color: 'var(--primary)',
+}));
+
+const IconWrapper = styled.span(() => ({
+  display: 'inline-block',
+  marginRight: layout.grid * 3,
+}));
+
+const EntityWrapper = styled.span(() => ({
```
stefcameron, line 61: Looks like you're using a `<span>` as a container, which isn't legal HTML. Seems like this should be a `<div>`.

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,343 @@
...
+    paddingTop: layout.pad * 1.5,
+    paddingBottom: layout.pad * 1.5,
+  },
+}));
+
+const ClusterLink = styled.a(() => ({
+  color: 'var(--primary)',
+}));
+
+const IconWrapper = styled.span(() => ({
```
stefcameron, line 56: Looks like you're using a `<span>` as a container, which isn't legal HTML. Seems like this should be a `<div>`.

stefcameron (changes requested):
Thanks for the Light theme screenshot. There seems to be some styling issue with the `<DrawerTitle>` component's background color that isn't switching with the theme here, but isn't present elsewhere in Lens. Notice how the background isn't visible in the Light theme, but it is in the Dark theme (I'm assuming that's because the background is there, and happens to match the Light theme's background gray color):

image

See how the `<DrawerTitle>` component behaves elsewhere (here in the `Catalog > Cluster > View Details` panel):

image

image

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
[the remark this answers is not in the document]
AntonNos: Strange, I fixed that. Maybe missclicked ctrl+z or so. Sorry, fixed now

src/strings.ts
AntonNos: Makes sense, fixed

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Done

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Makes sense, done.

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Yes, I used span because it was a table before. Fixed now, thanks for comment

stefcameron:
As discussed in Slack, let's address the light theme issue in a separate PR.

AntonNos:
> 

Yes, I pushed update and will work on it in separate PR

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,370 @@
...
+};
+
+export const SummaryPanel = ({ clusterEntity }) => {
+  //
+  // RENDER
+  //
+
+  const theme = document.body.classList.contains(lightThemeClassName)
+    ? lightTheme
+    : darkTheme;
```
stefcameron, line 168: This should go in #1174...

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,370 @@
...
+
+  const browserUrl = `${clusterEntity.metadata.cloudUrl}/projects/${clusterEntity.metadata.namespace}/clusters/${clusterEntity.metadata.name}`;
+
+  return (
+    <>
+      <DrawerTitleWrapper>
+        <PanelTitle isLightMode={theme.mode === themeModes.LIGHT}>
```
stefcameron, line 175: This should go in #1174...

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -0,0 +1,370 @@
...
+  color: ${({ isReady }) =>
+    isReady ? 'var(--colorSuccess)' : 'var(--textColorPrimary)'};
+`;
+
+const PanelTitle = styled(DrawerTitle)`
+  background-color: ${({ isLightMode }) =>
+    isLightMode
+      ? 'var(--layoutTabsLineColor)'
+      : 'var(--drawerSubtitleBackground)'};
+`;
```
stefcameron, line 90: This should go in #1174...

stefcameron (changes requested):
Nearly there! Just 3 small things that should be in #1174 instead of this PR since they're related to that fix, and the way they are here, it's not the right fix since it requires the page reload...

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Removed from current PR

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Removed from current PR

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Removed from current PR

PR #1174, opened 2022-11-02
[CCLEX-172] Theme provider for cluster page

AntonNos:


stefcameron:
👀

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -146,6 +155,8 @@ const getClusterLabelsCount = (clusterLabels) => {
 };
 
 export const SummaryPanel = ({ clusterEntity }) => {
+  const { theme } = useThemeSwitching();
```
stefcameron, line 158: At this level, you should be using `useTheme()` not this hook.

We need a new `ClusterPage.js` in the src/renderer/components/ClusterPage directory that is like the GlobalPage is now except that it has a `children: propTypes.node.isRequired` prop that requires a single child, which it then renders inside a `<ThemeProvider>`.

Then in the `renderer.tsx` where we have this:

```javascript
    {
      id: ROUTE_CLUSTER_OVERVIEW,
      components: {
        Page: () => <ClusterOverviewPage />,
      },
    },
```

Change it to this:

```javascript
    {
      id: ROUTE_CLUSTER_OVERVIEW,
      components: {
        Page: () => <ClusterPage><ClusterOverviewView /></ClusterPage>,
      },
    },
```

Would be good to then rename `ClusterOverviewPage` to `ClusterOverviewView` (and then `ClusterEventsView`, `ClusterDetailsView` also) so that we put a view _into_ the cluster page...

Finally, here in SummaryPanel, you don't need to get the theme here. You can get it in the PanelTitle styled component itself and do the right thing there:

```javascript
const PanelTitle = styled(DrawerTitle)(({ theme }) => ({
  backgroundColor: theme.mode === themeModes.LIGHT ? something : undefined,
  ...
}));
```

stefcameron (changes requested):
Good start, but I think we need to organize things a bit more, and need to make sure we're using Emotion correctly...

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Understand, thank you! Fixed

src/renderer/components/useThemeSwitching.js
```diff
@@ -0,0 +1,54 @@
+import { useEffect, useState } from 'react';
+import { lightThemeClassName, lightTheme, darkTheme } from './theme';
+
+export const useThemeSwitching = () => {
```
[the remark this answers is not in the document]
AntonNos: Done

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
```diff
@@ -76,6 +77,13 @@ const ServerStatus = styled.p`
     isReady ? 'var(--colorSuccess)' : 'var(--textColorPrimary)'};
 `;
 
+const PanelTitle = styled(DrawerTitle)(({ theme }) => ({
+  backgroundColor:
+    theme.mode === themeModes.LIGHT
+      ? 'var(--layoutTabsLineColor)'
+      : 'var(--drawerSubtitleBackground)',
+}));
```
stefcameron, line 85: Now that this still be the same thing in all views, let's pull it out to a separate little component in /src/renderer/components/ClusterPage and import it here as well as in the conditions view and details view...

src/renderer/components/useThemeSwitching.js
```diff
@@ -0,0 +1,60 @@
+import { useEffect, useState } from 'react';
+import { lightThemeClassName, lightTheme, darkTheme } from './theme';
+
+/**
+ * Meant for use in conjunction with Emotion's ThemeProvider
+ * in order to match the theme used to the theme Lens is using.
+ * Note: this is not a replacement for Emotion's useTheme() hook.
+ * @returns {object} Object with theme mode.
```
stefcameron, line 8: Would be clearer as `@returns {{ theme: Object }} Object with a `theme` property that is the theme object to be used by the ThemeProvider.`

stefcameron (changes requested):
Just two more improvements. Rest is looking great, and nice job on the additional tests!

src/renderer/components/useThemeSwitching.js
AntonNos: Thank you, fixed!

src/renderer/components/ClusterPage/Overview/SummaryPanel.js
AntonNos: Makes sense, done.

stefcameron (approved):
👍

PR #1176, opened 2022-11-03
[CCLEX-183]: Show '--' instead of the URL if StackLight is disabled

AntonNos:
### PR Checklist

### Updated Cluster details view
Screenshot 2022-11-03 at 16 32 10


stefcameron (approved):
👌 and bonus for adding the CHANGELOG entry! 😄

PR #1161, opened 2022-10-27
[CCLEX-176]: Details panel for Cluster page

AntonNos:
ezgif-3-bc4cc20330

Screenshot 2022-11-07 at 01 33 27

**FYI:** Tests for `ClusterDetailsPage` and rest of child components wasn't written, because need to modify `extension.js` (we are using `clusterEntity` from `Renderer.Catalog.catalogEntities` in `ClusterDetailsPage` and `DrawerTitle`/`DrawerItem` components in rest of child components).

PR #1177, opened 2022-11-03
[CCLEX-183] Handle not having lma; show enabled state in details

stefcameron:
The `lma` data of a cluster entity will be `null` if the cluster isn't configured with StackLight. The code needs to handle this case.

Also, now that we're not showing/hiding the LMA panel, it seems good to include an 'Enabled: Yes/No' field to indicate whether StackLight is configured on the cluster or not, since it might not be obvious just with the '--' values for URLs. For example, are all the URLs not available because SackLight is disabled, or because those services aren't available for some reason...?

### PR Checklist

    - Already added with #1176

### DIFF
--- a/src/api/types/Cluster.js
+++ b/src/api/types/Cluster.js
@@ -391,7 +391,7 @@ export class Cluster extends Node {
         ) {
           const { stacklight: stackLight = {} } =
             data.status.providerStatus.helm.releases;
-          return {
+          const lma = {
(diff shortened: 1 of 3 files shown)


stefcameron:
We'll need to do the same thing in #1161

stefcameron:
Waiting for a cluster to be ready that doesn't have StackLight enabled in order to test.

stefcameron:
Tested locally. Works!

PR #1195, opened 2022-11-15
[CCLEX-186] Add cluster events to cluster entities and fix bugs

stefcameron:
The main point of this change is to get cluster events into cluster entities in the Catalog so that we can display this data in the new Cluster Page.

But in doing this, I discovered we had a caching problem whenever we release a new version of the extension with new features that expect additional data to be in the Catalog entities generated (synced) by the extension: Unless the related API Resource changed versions (i.e. its `resourceVersion` was updated) in MCC, the entity would never get updated in Lens, and the additional data for the new features would never become available.

So now every Resource gets a new `cacheVersion` property, which ends-up in the model and then in the ensuing Catalog entity so we can compare not only the `resourceVersion` but also the `cacheVersion` when comparing enties to resources or models in the SyncManager.

Whenever Webpack is run, a new cache version is generated and embedded into the code. This will ensure that whenever we publish, the new package comes with a new cache version to cache-bust any existing data from previous versions out there the first time Lens runs after upgrading to the new extension package.

Also, for cluster events, these are `ClusterEvent` resources, and since they are technically full kube objects, they end-up as full entity models inside cluster entity models in `entity.spec.events` and so we need to compare those deeply when determining whether a cluster entity has changed or not in the SyncManager.

Finally, I moved some properties that were on leaf resource classes into common base classes to make managing them easier (e.g. Cluster and Machine had `conditions`, but these are common to the Node class so now its the Node class that provides the `conditions` property, as well as `toModel()` and `toString()` translations).

### PR Checklist


AntonNos:
👀

AntonNos (approved):
Looks great! I don't have anything to add😄

PR #1199, opened 2022-11-16
[CCLEX-186] Add cluster updates to API and cluster entities

stefcameron:
This adds an `updates` property on Cluster objects and cluster entities containing cluster/machine deployment/upgrade history, if available.

It also renames the 'Performance' tab to 'Updates' since we won't have a Performance tab anymore, but we're adding the Updates tab.

Since we need MCC v2.22 to get the new `stages[].status` field in `ClusterDeploymentStatus`, `ClusterUpgradeStatus`, `MachineDeploymentStatus`, and `MachineUpgradeStatus`, fetching updates is behind a new Webpack-enabled `FEAT_CLUSTER_PAGE_UPDATES_ENABLED` switch which can be set on the command line to opt in to seeing the Updates tab when running in Dev mode (or building Prod with the flag).

### PR Checklist

n/a (nothing new for the user, and no new data synced unless cluster page feature is enabled)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -78,7 +78,8 @@ const globals = {
   DEV_ENV: 'readonly',
   ENTITY_CACHE_VERSION: 'readonly',
   DEV_UNSAFE_NO_CERT: 'readonly',
-  FEAT_CLUSTER_PAGE_ENABLED: 'readonly', // TODO[clusterpage]: remove flag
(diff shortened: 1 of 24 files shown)


AntonNos:
👀

AntonNos (approved):
Nice work on this!

PR #1201, opened 2022-11-16
[CCLEX-187]: Added Namespace to entity details view and Cluster details page

AntonNos:
### PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - When setting the Management Cluster URL in the "Add Management Cluster" view, only the necessary portion of the URL will be used in order to establish the sync.
   - For example, if the user provides a URL to a child cluster, like `https://mcc.com/projects/my-project/clusters/my-cluster`, instead of the management cluster, only the origin, `https://mcc.com` in this case, will be used in order to help them be more successful when using the extension since the marketing term is somewhat confusing.
-- Added "Namespace" to each entity details view, replaced "Kind" with "Namespace" in Cluster details page.
+- Added "Namespace" to each entity details view in the Catalog.
 
 ## v5.2.1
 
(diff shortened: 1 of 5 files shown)


stefcameron:
👀

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - When setting the Management Cluster URL in the "Add Management Cluster" view, only the necessary portion of the URL will be used in order to establish the sync.
   - For example, if the user provides a URL to a child cluster, like `https://mcc.com/projects/my-project/clusters/my-cluster`, instead of the management cluster, only the origin, `https://mcc.com` in this case, will be used in order to help them be more successful when using the extension since the marketing term is somewhat confusing.
+- Added "Namespace" to each entity details view, replaced "Kind" with "Namespace" in Cluster details page.
```
stefcameron, line 7: Since we haven't published the cluster page yet, only this is necessary:
```suggestion
- Added "Namespace" to each entity details view in the Catalog.
```

src/strings.ts
```diff
@@ -213,6 +213,7 @@ export const catalog: Dict = {
         props: {
           dateCreated: () => 'Date created',
           lastSync: () => 'Last sync',
+          namespace: () => 'Namespace',
```
stefcameron, line 216: I forgot that we call a namespace a "project" in MCC. Let's use `'Project'` instead of `'Namespace'`

```suggestion
          namespace: () => 'Project',
```

stefcameron (changes requested):
In the "Cluster Page > Overview", we're using `clusterName: () => 'Cluster name',` -- let's just make it `clusterName: () => 'Name',` to be consistent with just using `'Name'` on the Details panel. Also, let's move the namespace up to be right after the cluster name in the Overview panel.

CHANGELOG.md
AntonNos: Fixed

src/strings.ts
AntonNos: Fixed

AntonNos:
> In the "Cluster Page > Overview", we're using `clusterName: () => 'Cluster name',` -- let's just make it `clusterName: () => 'Name',` to be consistent with just using `'Name'` on the Details panel. Also, let's move the namespace up to be right after the cluster name in the Overview panel.

Fixed

stefcameron (approved):
👌

PR #1224, opened 2022-12-12
[CCLEX-185]: History panel

AntonNos:
`EventsTable` and `HistoryTable` components were moved to re-usable `ItemsTable` component. Also did some changes in `EventsPanel` to meet `ItemsTable` structure.

### **Mocked data to see, how statuses are looking:**
Screenshot 2022-12-11 at 23 15 52

### **Real data:**
Screenshot 2022-12-12 at 03 04 38

### DIFF
--- a/src/renderer/components/ClusterPage/Events/EventsPanel.js
+++ b/src/renderer/components/ClusterPage/Events/EventsPanel.js
@@ -14,7 +14,6 @@ import { formatDate } from '../../../rendererUtil';
 import { apiKinds } from '../../../../api/apiConstants';
 import { useTableSearch } from '../useTableSearch';
 import { useCloudSync } from '../useCloudSync';
-import { handleCloudSync } from '../clusterPageUtil';
 import { ItemsTable } from '../ItemsTable';
 import {
   TablePanelWrapper,
@@ -90,7 +89,7 @@ const tableHeaders = [
 /**
  * Creates array with arrays of events values objects for future render.
  * @param {Array} events array with events update objects.
- * @returns {{ Array }} array with arrays of objects.
+ * @returns {Array<Array<{ text: string, color?: string, isBiggerCell?: boolean }>>} array with arrays of objects.
  */
 const generateItems = (events) => {
   return events.map((event) => {
@@ -119,7 +118,7 @@ const generateItems = (events) => {
             : strings.clusterPage.common.emptyValue(),
       },
       {
-        text: event.spec.count || unknownValue(),
+        text: `${event.spec.count}` || unknownValue(),
       },
     ];
   });
@@ -143,7 +142,7 @@ export const EventsPanel = ({ clusterEntity }) => {
     searchItems: clusterEntity.spec.events,
   });
 
-  const { isCloudFetching, cloudStatus } = useCloudSync(
+  const { isCloudFetching, cloudStatus, syncCloud } = useCloudSync(
     clusterEntity.metadata.cloudUrl
   );
 
@@ -221,11 +220,6 @@ export const EventsPanel = ({ clusterEntity }) => {
     [clusterEntity.spec.events]
   );
 
-  const syncCloud = useCallback(
-    () => handleCloudSync(clusterEntity.metadata.cloudUrl),
-    [clusterEntity.metadata.cloudUrl]
-  );
-
   const handleSelectChange = useCallback(
     (newSelection) => {
       const newValue = newSelection?.value || null;
@@ -271,7 +265,7 @@ export const EventsPanel = ({ clusterEntity }) => {
               isCloudFetching || cloudStatus !== CONNECTION_STATUSES.CONNECTED
             }
             isCloudFetching={isCloudFetching}
-            onClick={syncCloud}
+            onClick={() => syncCloud()}
           >
             <Icon material="refresh" />
           </TableSyncButton>
--- a/src/renderer/components/ClusterPage/Events/EventsPanel.js
+++ b/src/renderer/components/ClusterPage/Events/EventsPanel.js
@@ -83,6 +83,11 @@ const tableHeaders = [
   },
 ];
 
+/**
+ * Creates array with arrays of events values objects for future render.
+ * @param {Array} events array with events update objects.
+ * @returns {{ Array }} array with arrays of objects.
+ */
 const generateItems = (events) => {
   return events.map((event) => {
     return [
--- a/src/renderer/components/ClusterPage/Events/EventsPanel.js
+++ b/src/renderer/components/ClusterPage/Events/EventsPanel.js
@@ -7,10 +7,7 @@ import {
   useCallback,
   useMemo,
 } from 'react';
-import styled from '@emotion/styled';
-import { keyframes } from '@emotion/css';
 import { Renderer } from '@k8slens/extensions';
-import { layout } from '../../styles';
 import * as strings from '../../../../strings';
 import { CONNECTION_STATUSES } from '../../../../common/Cloud';
 import { formatDate } from '../../../rendererUtil';
@@ -19,6 +16,13 @@ import { useTableSearch } from '../useTableSearch';
 import { useCloudSync } from '../useCloudSync';
 import { handleCloudSync } from '../clusterPageUtil';
 import { ItemsTable } from '../ItemsTable';
+import {
+  TablePanelWrapper,
+  TableTopItems,
+  TableSettings,
+  TableSyncButton,
+  TableSearch,
+} from '../clusterPageComponents';
 
 const TABLE_HEADER_IDS = {
   TYPE: 'type',
@@ -30,7 +34,7 @@ const TABLE_HEADER_IDS = {
 };
 
 const {
-  Component: { Icon, SearchInput, Select },
+  Component: { Icon, Select },
 } = Renderer;
 
 const {
@@ -120,48 +124,6 @@ const generateItems = (events) => {
     ];
   });
 };
-
-//
-// INTERNAL STYLED COMPONENTS
-//
-
-const rotate = keyframes`
-  from {
-    transform: rotate(0deg);
-  }
-  to {
-    transform: rotate(360deg);
-  }
-`;
-
-const PanelWrapper = styled.div(() => ({
-  background: 'var(--contentColor)',
-  height: '100%',
-}));
-
-const TopItems = styled.div(() => ({
-  display: 'flex',
-  justifyContent: 'space-between',
-  alignItems: 'center',
-  padding: layout.pad * 2,
-}));
-
-const Settings = styled.div(() => ({
-  display: 'flex',
-  alignItems: 'center',
-}));
-
-const SyncButton = styled.button`
-  margin-right: ${layout.pad * 2}px;
-  pointer-events: ${({ isDisabled }) => (isDisabled ? 'none' : 'auto')};
-  opacity: ${({ isDisabled }) => (isDisabled ? '0.5' : '1')};
-  animation: ${({ isCloudFetching }) =>
-    isCloudFetching ? `${rotate} 2s linear infinite` : 'none'};
-`;
-
-const Search = styled(SearchInput)(() => ({
-  marginLeft: layout.pad * 1.25,
-}));
 
 //
 // MAIN COMPONENT
@@ -297,14 +259,14 @@ export const EventsPanel = ({ clusterEntity }) => {
   }, [filters]);
 
   return (
-    <PanelWrapper>
-      <TopItems ref={targetRef}>
+    <TablePanelWrapper>
+      <TableTopItems ref={targetRef}>
         <p>{strings.clusterPage.pages.events.title()}</p>
         <p>
           {strings.clusterPage.pages.events.itemsAmount(filteredEvents.length)}
         </p>
-        <Settings>
-          <SyncButton
+        <TableSettings>
+          <TableSyncButton
             isDisabled={
               isCloudFetching || cloudStatus !== CONNECTION_STATUSES.CONNECTED
             }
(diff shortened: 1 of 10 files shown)


src/renderer/components/ClusterPage/History/HistoryPanel.js
```diff
@@ -0,0 +1,323 @@
...
+        message: stage.message,
+        targetKind: item.spec.targetKind,
+        targetName: item.spec.targetName,
+        fromRelease: item.spec.fromRelease,
+        release: item.spec.release,
+      });
+    });
+  });
+  return arr;
+};
```
stefcameron, line 168: I know `map()` iterates, but you're really just iterating the lists without transforming them. Using `forEach()` in both cases would be more accurate and make the code easier to understand.

But I suggest you use `flatMap()` to "map" the `history` array to an array of items. I also think you should rename the `history` parameter to `updates` because at the point the `updates` list is received, it's an `Array<ResourceUpdate>` -- it's not what you call "history" items yet. That's the _output_ of this function.

```javascript
return updates.flatMap((update) =>
  update.spec.stages.map((stage) => ({
    status: stage.status,
    timeAt: stage.timeAt,
    name: stage.name,
    message: stage.message,
    targetKind: update.spec.targetKind,
    targetName: update.spec.targetName,
    fromRelease: update.spec.fromRelease,
    release: update.spec.release,
  })));
```

I think the above is a lot more clear on what a "history" object is (a mix of stage and related update data). WDYT?

The types in the JSDoc should also be clarified:

```
 * @param {Array<ResourceUpdate>} updates List of resource updates.
 * @returns {Array<{ status: string, timeAt: string, name: string, message: string, targetKind: string, targetName: string, fromRelease: string|null, release: string }>} List of history objects.
```

Also renaming the function to `getHistory = (updates) => ...` would make the code more clear since this function is transforming a list of resource updates into a list of history.

src/renderer/components/ClusterPage/ItemsTable.js
```diff
@@ -46,27 +43,29 @@ const ResetFiltersButton = styled.button(() => ({
   borderBottom: '1px dotted',
 }));
 
-const TableCellWithPadding = styled(TableCell)(() => ({
+const TableCellWithPadding = styled(TableCell)((props) => ({
   paddingTop: layout.grid * 2.5,
   paddingBottom: layout.grid * 2.5,
+  wordBreak: 'normal',
+
+  '&&&': {
```
stefcameron, line 51: Triple `&&&`? I believe just a single `&` will suffice.

src/renderer/components/ClusterPage/ItemsTable.js
```diff
@@ -121,73 +127,52 @@ export const EventsTable = ({
...
+      ) : items.length > 0 ? (
+        items.map((item, rowIndex) => (
+          <TableRow key={rowIndex}>
+            {item.map((singleItem, itemIndex) => (
+              <TableCellWithPadding
+                key={itemIndex}
+                isBiggerCell={singleItem.isBiggerCell}
+                style={{
+                  color: singleItem.color || 'var(--textColorPrimary)',
+                }}
```
stefcameron, line 142: It's a bit strange to be setting a `style` on a styled component. Why not pass the `singleItem` to the `TableCellWithPadding` and have it do this internally when it sets the color style?

src/renderer/components/ClusterPage/ItemsTable.js
```diff
@@ -121,73 +127,52 @@ export const EventsTable = ({
...
-EventsTable.propTypes = {
+ItemsTable.propTypes = {
   tableHeaders: propTypes.arrayOf(
     propTypes.shape({
       id: propTypes.string,
       label: propTypes.string,
     })
   ).isRequired,
-  events: propTypes.arrayOf(propTypes.object),
+  items: propTypes.arrayOf(propTypes.array),
```
stefcameron, line 175: Can you use `propTypes.arrayOf(propTypes.shape({ ... }))` and explicitly describe what those items look like? That will be really helpful in the future when coming back to this code.

src/renderer/components/ClusterPage/clusterPageComponents.js
```diff
@@ -31,3 +37,41 @@ export const DrawerItemsWrapper = styled.div(() => ({
...
+  alignItems: 'center',
+}));
+
+export const TableSyncButton = styled.button`
+  margin-right: ${layout.pad * 2}px;
+  pointer-events: ${({ isDisabled }) => (isDisabled ? 'none' : 'auto')};
+  opacity: ${({ isDisabled }) => (isDisabled ? '0.5' : '1')};
+  animation: ${({ isCloudFetching }) =>
+    isCloudFetching ? `${rotate} 2s linear infinite` : 'none'};
+`;
```
stefcameron, line 73: Since there are many props expected, it would be better to use the object for instead of the template string form so that the expected props can be easier to see:

```javascript
export const TableSyncButton = styled.button(({ isDisabled, isCloudFetching }) => ({ ... }));
```

src/renderer/components/ClusterPage/clusterPageUtil.js
```diff
@@ -88,3 +89,11 @@ export const getProvider = (provider) => {
...
   }
 };
+
+/**
+ * Starts cloud sync.
+ * @param {string} cloudUrl Cloud url to sync.
+ */
+export const handleCloudSync = (cloudUrl) => {
+  IpcRenderer.getInstance().invoke(consts.ipcEvents.invoke.SYNC_NOW, cloudUrl);
+};
```
stefcameron, line 99: This feels like something the `useCloudSync()` hook should return. Feels a bit strange having this be here, separate from the hook.

stefcameron (changes requested):
Nice work reusing the ItemsTable!

src/renderer/components/ClusterPage/History/HistoryPanel.js
```diff
@@ -0,0 +1,323 @@
...
+      return 'var(--colorError)';
+    default:
+      return 'var(--textColorPrimary)';
+  }
+};
+
+/**
+ * Creates array with arrays of history values objects for future render.
+ * @param {Array} history array with history update objects.
+ * @returns {{ Array }} array with arrays of objects.
```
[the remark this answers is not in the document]
AntonNos: Fixed

src/renderer/components/ClusterPage/History/HistoryPanel.js
AntonNos: You are right, it looks much cleaner. Thank you. Fixed

src/renderer/components/ClusterPage/ItemsTable.js
AntonNos: Yes, I used triple `&&&` here, because I need to override styles with higher specificity. According to docs, need to use triple ampersands. It doesn't work with single one. Here is docs - https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity.

src/renderer/components/ClusterPage/ItemsTable.js
AntonNos: Right, thank you. Fixed

src/renderer/components/ClusterPage/ItemsTable.js
AntonNos: Done

src/renderer/components/ClusterPage/History/HistoryPanel.js
```diff
@@ -0,0 +1,323 @@
...
+  MACHINE: 'machine',
+  FROM_RELEASE: 'from-release',
+  TO_RELEASE: 'to-release',
+};
+
+const HISTORY_STATUSES = {
+  IN_PROGRESS: 'in progress',
+  SUCCESS: 'success',
+  FAILED: 'failed',
+};
```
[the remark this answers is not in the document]
AntonNos: Fixed

src/renderer/components/ClusterPage/clusterPageComponents.js
AntonNos: Makes sense, fixed

src/renderer/components/ClusterPage/clusterPageUtil.js
AntonNos: You are right! I don't know why I placed it outside of hook🤷‍♂️
Thanks for tip

src/renderer/components/ClusterPage/History/HistoryPanel.js
```diff
@@ -94,11 +87,11 @@ const tableHeaders = [
...
   switch (status.toLowerCase()) {
-    case HISTORY_STATUSES.IN_PROGRESS:
+    case apiUpdateStatuses.IN_PROGRESS.toLowerCase():
       return 'var(--colorWarning)';
-    case HISTORY_STATUSES.SUCCESS:
+    case apiUpdateStatuses.SUCCESS.toLowerCase():
       return 'var(--colorSuccess)';
-    case HISTORY_STATUSES.FAILED:
+    case apiUpdateStatuses.FAILED.toLowerCase():
       return 'var(--colorError)';
```
stefcameron, line 95: Is it really necessary to lowercase everything? The `status` is supposed to be a specific string we can count on. I think we can drop the lowercasing.

src/renderer/components/ClusterPage/History/HistoryPanel.js
```diff
@@ -291,7 +275,7 @@ export const HistoryPanel = ({ clusterEntity }) => {
               isCloudFetching || cloudStatus !== CONNECTION_STATUSES.CONNECTED
             }
             isCloudFetching={isCloudFetching}
-            onClick={syncCloud}
+            onClick={() => syncCloud()}
```
stefcameron, line 278: Same here, we've lost the cached callback...

src/renderer/components/ClusterPage/useCloudSync.js
```diff
@@ -46,5 +48,12 @@ export const useCloudSync = (cloudUrl) => {
     };
   }, [clouds, cloudUrl]);
 
-  return { isCloudFetching, cloudStatus };
+  const syncCloud = () => {
```
stefcameron, line 51: Since this is in a hook, I believe it's legal to do this:
```suggestion
  const handleSyncCloud = useCallback(() => {...}, cloudUrl);
```

And then return `handleSyncCloud` as the handler to use.

stefcameron (commented):
Looking good! Just a couple more small things.

src/renderer/components/ClusterPage/ItemsTable.js
stefcameron: Huh, that's cool. I didn't know about this. Thanks!

src/renderer/components/ClusterPage/History/HistoryPanel.js
AntonNos: You are right, we can remove it. Fixed

src/renderer/components/ClusterPage/useCloudSync.js
AntonNos: Good idea, done.

src/renderer/components/ClusterPage/History/HistoryPanel.js
AntonNos: Fixed

src/renderer/components/ClusterPage/Events/EventsPanel.js
```diff
@@ -271,7 +265,7 @@ export const EventsPanel = ({ clusterEntity }) => {
               isCloudFetching || cloudStatus !== CONNECTION_STATUSES.CONNECTED
             }
             isCloudFetching={isCloudFetching}
-            onClick={syncCloud}
+            onClick={() => syncCloud()}
```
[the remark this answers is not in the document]
AntonNos: Fixed

src/renderer/components/ClusterPage/useCloudSync.js
```diff
@@ -48,12 +48,13 @@ export const useCloudSync = (cloudUrl) => {
...
-      consts.ipcEvents.invoke.SYNC_NOW,
-      cloudUrl
-    );
+  const syncCloud = (url) => {
+    IpcRenderer.getInstance().invoke(consts.ipcEvents.invoke.SYNC_NOW, url);
   };
 
-  return { isCloudFetching, cloudStatus, syncCloud };
+  const handleSyncCloud = useCallback(() => {
+    syncCloud(cloudUrl);
```
stefcameron, line 56: I'm surprised the ESLint "rules of hooks" _exhaustive dependencies_ rule isn't flagging this as a missing dependency in the `useCallback`'s dependency list because as it's currently setup, the callback will change only when `cloudUrl` changes, yet the callback calls an ever-changing `syncCloud()` function (unnecessarily changes on every render when the hook is called).

`syncCloud()` is just one line and not called from anywhere else. I suggest either getting rid of it and having the `IpcRenderer` call right here, or moving the `syncCloud()` function _outside_ the `useCloudSync` hook.

AntonNos: I moved `IpcRenderer` call inside `useCallback` and get rid of `syncCloud`.

stefcameron:
@AntonNos Thanks for addressing all my comments! Just one last thing about the layout:

image

I think we need to give more room for the Name column. I'm thinking equal "weight" in the flex layout as we give to the Message, because the Message is only used for details. Otherwise, they use the Name as the main description of the history update.

AntonNos:
> @AntonNos Thanks for addressing all my comments! Just one last thing about the layout:
> 
> image
> 
> I think we need to give more room for the Name column. I'm thinking equal "weight" in the flex layout as we give to the Message, because the Message is only used for details. Otherwise, they use the Name as the main description of the history update.

Fixed width of `name` column, also made small changes in render headers in `ItemsTable` component.
Screenshot 2022-12-13 at 21 23 42

stefcameron (approved):
Thanks for adjusting the column widths. Looks great, nice work!

PR #1244, opened 2023-01-04
[CCLEX-173]: Health panel in Cluster page

AntonNos:
ezgif-2-629ab89fea

I made 3 fullwidth columns, because only on >1920px monitors it looks good with 50% columns.

**Update:** 

Light mode:
Screenshot 2023-01-12 at 21 25 23

No lma:
Screenshot 2023-01-13 at 02 05 01

**UPD: 50% width columns in Overview panel** 
Screenshot 2023-01-18 at 00 42 36


stefcameron:
👀

src/renderer/components/ClusterPage/Overview/MetricChart.js
```diff
@@ -0,0 +1,73 @@
+import propTypes from 'prop-types';
+import { useEffect, useState } from 'react';
+import { Chart as ChartJS, ArcElement } from 'chart.js';
+import { Doughnut } from 'react-chartjs-2';
+
+ChartJS.register(ArcElement);
+
+const style = getComputedStyle(document.body);
```
stefcameron, line 8: I don't think you can do this just once when the module is loaded because I don't think that will work with theming. You need to do this every time something gets rendered, which is too bad because calling `getComputedStyle()` causes [layout reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#what-forces-layout--reflow), which means additional renders. Generally, it's best to avoid these APIs as much as possible. But I understand why you're using it: To avoid hard-coding color HEX values. It's probably fine since we're not calling it in a loop, which would be the worst case scenario.

But still, I think you need to call it _inside_ `getData()`, and `defaultData` should be a function `getDefaultData()` that calls it inside too, and returns the default data object...

src/renderer/components/ClusterPage/Overview/SingleMetric.js
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <>
+      <div>
+        <ChartWrapper>
+          <MetricChart
+            chartColor={chartColor}
+            chartFillPercentage={chartFillPercentage}
+          />
+          <PercentageValue color={chartColor}>
+            {chartFillPercentage}%
```
stefcameron, line 56: I know it's just a percentage sign, but I'd prefer to keep all formatted strings coming from `strings.ts` as much as possible, so a strings.ts string that takes the `chartFillPercentage` and returns `"${chartFillPercentage}%"` would be my preference over hard coding the "%" string here.

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+ * @param {string|number} obj.system System value from CPU metrics.
+ * @param {string|number} obj.io I/O value from CPU metrics.
+ * @param {string|number} obj.idle Idle value from CPU metrics.
+ * @returns {Array<{ 'Used': string, 'System': string, 'I/O': string, 'Idle': string }>} array with objects of CPU metrics.
+ */
+const getCpuData = ({ used, system, io, idle }) => {
+  return [
+    {
+      [`${strings.clusterPage.pages.overview.health.metrics.cpu.used()}`]: used
+        ? `${Math.round(used * 100)}%`
```
stefcameron, line 63: I know it's just a percentage sign, but I'd prefer to keep all formatted strings coming from `strings.ts` as much as possible, so a strings.ts string that takes the `chartFillPercentage` and returns `"${chartFillPercentage}%"` would be my preference over hard coding the "%" string here.

package.json
```diff
@@ -61,7 +61,10 @@
     "download:kubelogin:license:nix": "yarn create:bin:nix && ([[ -f ./bin/kubelogin-license.txt ]] || curl -slL https://raw.githubusercontent.com/int128/kubelogin/v1.25.3/LICENSE > ./bin/kubelogin-license.txt)",
     "download:kubelogin:license:windows": "yarn create:bin:windows && sh -c \"[[ -f ./bin/kubelogin-license.txt ]] || curl -slL https://raw.githubusercontent.com/int128/kubelogin/v1.25.3/LICENSE > ./bin/kubelogin-license.txt\""
   },
-  "dependencies": {},
+  "dependencies": {
+    "chart.js": "^4.1.1",
+    "react-chartjs-2": "^5.1.0"
```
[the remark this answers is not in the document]
AntonNos: Done

src/renderer/components/ClusterPage/Overview/ClusterOverviewView.js
```diff
@@ -21,19 +21,25 @@ export const ClusterOverviewView = function ({ clusterEntity }) {
...
-      <ConditionsPanel clusterEntity={clusterEntity} />
+      <div>
+        <div>
+          <SummaryPanel clusterEntity={clusterEntity} />
+        </div>
+        <div>
+          {FEAT_CLUSTER_PAGE_HEALTH_ENABLED ? (
+            <HealthPanel clusterEntity={clusterEntity} />
+          ) : null}
+        </div>
```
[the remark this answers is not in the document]
AntonNos: Right, thanks. Fixed

src/renderer/components/ClusterPage/Overview/ClusterOverviewView.js
```diff
@@ -21,19 +21,25 @@ export const ClusterOverviewView = function ({ clusterEntity }) {
...
+        <div>
+          <DrawerTitleWrapper>
+            <PanelTitle
+              title={strings.clusterPage.pages.overview.clusterConditions.title()}
+            />
+          </DrawerTitleWrapper>
```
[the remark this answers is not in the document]
AntonNos: Hmm, I don't know why title is here😄 Fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+import { MetricTitle } from './MetricTitle';
+import { layout } from '../../styles';
+import { logger } from '../../../../util/logger';
+import * as strings from '../../../../strings';
+
+const UPDATE_METRICS_INTERVAL = 60000; // 60000ms = 1min
+
+/**
+ * Converts size in bytes to KB, MB, GB etc.
+ * @param {string|number} bytes The size in bytes.
```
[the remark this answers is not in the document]
AntonNos: Yes, this is much cleaner, thanks. Fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+  return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`;
+};
+
+/**
+ * Creates array with objects of CPU metrics for future render.
+ * @param {Object} obj CPU metrics values.
+ * @param {string|number} obj.used Used value from CPU metrics.
+ * @param {string|number} obj.system System value from CPU metrics.
+ * @param {string|number} obj.io I/O value from CPU metrics.
+ * @param {string|number} obj.idle Idle value from CPU metrics.
```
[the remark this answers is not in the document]
AntonNos: Fixed, based on our discussion😉

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+    {
+      [`${strings.clusterPage.pages.overview.health.metrics.cpu.io()}`]: io
+        ? `${Math.round(io * 100)}%`
+        : strings.catalog.entities.common.details.emptyValue(),
+    },
+    {
+      [`${strings.clusterPage.pages.overview.health.metrics.cpu.idle()}`]: idle
+        ? `${Math.round(idle * 100)}%`
+        : strings.catalog.entities.common.details.emptyValue(),
+    },
```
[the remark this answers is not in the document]
AntonNos: Makes sense. Replaced the check from `used ? '' : ''` to `used >= 0 ? '' : ''`

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+};
+
+/**
+ * Creates array with objects of CPU metrics for future render.
+ * @param {Object} obj CPU metrics values.
+ * @param {string|number} obj.used Used value from CPU metrics.
+ * @param {string|number} obj.system System value from CPU metrics.
+ * @param {string|number} obj.io I/O value from CPU metrics.
+ * @param {string|number} obj.idle Idle value from CPU metrics.
+ * @returns {Array<{ 'Used': string, 'System': string, 'I/O': string, 'Idle': string }>} array with objects of CPU metrics.
```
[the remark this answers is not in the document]
AntonNos: You are right, it will look much cleaner. Fixed.
Thanks for tip!

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+  ];
+};
+
+/**
+ * Creates array with objects of Memory metrics for future render.
+ * @param {Object} obj Memory metrics values.
+ * @param {string|number} obj.available Available memory value from Memory metrics.
+ * @param {string|number} obj.capacity Capacity value from Memory metrics.
+ * @param {string|number} obj.allocated Allocated memory value from Memory metrics.
+ * @returns {Array<{ 'Available': string, 'Capacity': string, 'Allocated': string }>} array with objects of Memory metrics.
```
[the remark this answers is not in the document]
AntonNos: Fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+  ];
+};
+
+/**
+ * Creates array with objects of Storage metrics for future render.
+ * @param {Object} obj Storage metrics values.
+ * @param {string|number} obj.used Used storage value from Storage metrics.
+ * @param {string|number} obj.capacity Capacity value from Storage metrics.
+ * @param {string|number} obj.available Available storage value from Storage metrics.
+ * @returns {Array<{ 'Used': string, 'Capacity': string, 'Available': string }>} array with objects of Storage metrics.
```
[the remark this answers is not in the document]
AntonNos: Fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+  const isStorageFirstRender = useRef(true);
+
+  useEffect(() => {
+    /**
+     * Gets CPU, Memory and Storage metrics from Prometheus.
+     * @param {Object} cloud Cloud.
+     */
+    const getMetrics = async (cloud) => {
+      const promUrl =
+        'https://a3f5c7cc1fc794891a8e86d6f29a8fe7-616423632.us-east-2.elb.amazonaws.com';
```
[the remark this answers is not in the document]
AntonNos: Ohh, I thought that this one is mocked. Fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+  const [memoryData, setMemoryData] = useState([]);
+  const [storageData, setStorageData] = useState([]);
+  const [cpuPercentage, setCpuPercentage] = useState(0);
+  const [memoryPercentage, setMemoryPercentage] = useState(0);
+  const [storagePercentage, setStoragePercentage] = useState(0);
+  const [timerTrigger, setTimerTrigger] = useState(0);
+
+  const isCpuFirstRender = useRef(true);
+  const isMemoryFirstRender = useRef(true);
+  const isStorageFirstRender = useRef(true);
```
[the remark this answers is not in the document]
AntonNos: You are right. We can use that case, when metrics are `null`. Thanks for tip! Fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+        setStorageMetrics(metrics);
+      } catch (err) {
+        logger.error(err.message);
+        setStorageMetrics({});
+      }
+
+      // Changes timer trigger every minute to update metrics.
+      setTimeout(() => {
+        setTimerTrigger(timerTrigger + 1);
+      }, UPDATE_METRICS_INTERVAL);
```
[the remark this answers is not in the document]
AntonNos: Done

src/renderer/components/ClusterPage/Overview/SingleMetric.js
AntonNos: Makes sense, fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: Fixed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,331 @@
...
+const UPDATE_METRICS_INTERVAL = 60000; // 60000ms = 1min
+
+/**
+ * Converts size in bytes to KB, MB, GB etc.
+ * @param {string|number} bytes The size in bytes.
+ * @param {number} decimals The number of characters after the period.
+ * @returns {string} `--` if no bytes provided, formatted size in other case.
+ */
+const formatBytes = (bytes, decimals = 2) => {
+  if (!bytes) {
```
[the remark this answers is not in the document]
AntonNos: Fixed

stefcameron:
👀

src/renderer/components/ClusterPage/clusterPageComponents.js
```diff
@@ -76,3 +76,11 @@ export const TableSyncButton = styled.button(
 export const TableSearch = styled(SearchInput)(() => ({
   marginLeft: layout.pad * 1.25,
 }));
+
+export const PanelsWrapper = styled.div(() => ({
+  overflow: 'auto',
```
stefcameron, line 81: Yesterday, I fixed an issue with scrolling by putting `overflow: 'auto',` on the PageContainer component. Should it be removed in light of this style here, or should this one be removed in favor of the one I added?

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,360 @@
...
+const getCpuData = ({ used, system, io, idle }) => {
+  return [
+    {
+      label: strings.clusterPage.pages.overview.health.metrics.cpu.used(),
+      value:
+        typeof used === 'number' && !Number.isNaN(used)
+          ? strings.clusterPage.pages.overview.health.metrics.cpu.percentageValue(
+              Math.round(used * 100)
+            )
+          : strings.catalog.entities.common.details.emptyValue(),
```
stefcameron, line 68: Since all 4 metrics are subject to the same checks and formatting, could this be moved into a small `formatPct()` function similar to `formatBytes()` that takes a `pct` value and does the checks and formatting? Then there's only one place to tweak if changes are needed, and the rest of the code just becomes like
```
{
   label: strings...,
   value: formatPct(system),
}
...
```

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,360 @@
...
+    };
+  }, [updateMetricsTimeoutId]);
+
+  useEffect(() => {
+    /**
+     * Gets CPU, Memory and Storage metrics from Prometheus.
+     * @param {Object} cloud Cloud.
+     */
+    const getMetrics = async (cloud) => {
+      const promUrl = clusterEntity.spec.lma.prometheusUrl;
```
stefcameron, line 207: This is good, but we will have a case where `lma === null/undefined` because StackLight isn't installed, or also (maybe, just covering all possibilities) where `lma.prometheusUrl === null/undefined` because Prometheus isn't there.

In that case, we should be in this case: https://www.figma.com/file/6LYlfoeqjDMEdtKnhmoLgd/MCC%2FLens?node-id=56%3A2194&t=l8ZhpR6gvU80ZgbJ-4

image

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,360 @@
...
+          setMemoryMetrics({});
+        }
+
+        try {
+          const metrics = await getDiskMetrics(cloud, promUrl);
+          setStorageMetrics(metrics);
+        } catch (err) {
+          logger.error(err.message);
+          setStorageMetrics({});
+        }
```
stefcameron, line 232: Instead of executing these requests serially, it would be faster to do them in parallel with `Promise.allSettled()` (which will give you results for each whether they succeed or fail so you can still show individual error messages).

And you should really test the error cases. `logger.error()` expects a "scope" as its first parameter, and then a message as its second, and will throw if it only gets one parameter... Busted! 😉 

Since this is running inside an effect, you have to get a bit creative. This is what I would suggest. The point is using a scope that would help you quickly identify _where_ this log message came from so you can quickly find the spot where the error occurred.

```javascript
import { logger, logValue } from '../util/logger';
...
logger.error(
  'HealthPanel.useEffect.getMetrics()',
  `Failed to get CPU metrics; error=${logValue(err)}`
);
```

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,360 @@
...
+        setStorageMetrics({});
+      }
+    };
+
+    getMetrics(clouds[clusterEntity.metadata.cloudUrl]);
+  }, [clouds, clusterEntity, timerTrigger]);
+
+  useEffect(() => {
+    if (cpuMetrics !== null) {
+      if (cpuMetrics) {
```
stefcameron, line 252: Since the initial state of `cpuMetrics` is `null`, I'm struggling to find a case where the `else` statement on line 265 would ever get called. First pass, `cpuMetrics === null` so this effect does nothing. Meanwhile, the effect above is executing and waiting for the async request to get CPU metrics to complete. Request completes, whether successful or not, we have `cpuMetrics !== null`. Then this effect runs again because `cpuMetrics` has changed, and then both lines 251 and 252 succeed and we fall into the `if` block.

Also, if that `else` block were ever to be called, it would throw because `getCpuData()`'s signature is `getCpuData({ ... })` which means it expects the first parameter to be an object it can destructure, yet you're passing `undefined`. That will cause an exception. You'd had to define `getCpuData` as `getCpuData({ ... } = {})` to allow it to be called as `getCpuData()` or `getCpuData({})` or `getCpuData({ ... })`.

But I don't think that's necessary, because the effect above that makes the async calls handles failures and will always set `cpuMetrics` to _something_ that isn't `null`.

So I think you can eliminate line 251 as well as the `else` block at line 264.

And the same goes for memory and storage effects below.

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,360 @@
...
+  const [storageData, setStorageData] = useState([]);
+  const [cpuPercentage, setCpuPercentage] = useState(0);
+  const [memoryPercentage, setMemoryPercentage] = useState(0);
+  const [storagePercentage, setStoragePercentage] = useState(0);
+  const [updateMetricsTimeoutId, setUpdateMetricsTimeoutId] = useState(null);
+  const [timerTrigger, setTimerTrigger] = useState(0);
+
+  // Canceling the timeout if the component gets unmounted
+  useEffect(() => {
+    mounted.current = true;
```
stefcameron, line 193: What is this ref used for?

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -1,7 +1,360 @@
...
+
+    return () => {
+      clearTimeout(updateMetricsTimeoutId);
+      mounted.current = false;
+    };
```
stefcameron, line 198: I was expecting this to be in the effect where you're making the async calls and setting the timeout. You _probably_ don't need to keep the timer ID in state either because you're already using another state variable (`timerTrigger`) to cause the timeout to be re-established when it expires.

src/strings.ts
```diff
@@ -382,6 +382,64 @@ export const clusterPage: Dict = {
...
+          storage: {
+            title: () => 'Storage',
+            used: () => 'Used',
+            capacity: () => 'Capacity',
+            available: () => 'Available',
+            tooltipInfoHtml: () => `
+              <p>Metrics are received from the kubernetes cluster and may differ from OS level metrics.</p>
+              <p>Capacity: Total disk space across.</p>
+              <p>Available: Total free disk space.</p>
+              <p>Used: Total consumed disk space.</p>
```
stefcameron, line 424: These strings are all missing "across the cluster".

stefcameron, line 424: Actually, most of these strings aren't what is in https://mirantis.jira.com/browse/CCLEX-173 at the "Metric info explanations" section. 🤔

stefcameron (changes requested):
Thank you for those improvements!

stefcameron:
> I made 3 fullwidth columns, because only on >1920px monitors it looks good with 50% columns.

Really? It's not possible to make the graphs smaller (and so narrower) to give in a side-by-side panel with Overview?

Also, you know I'm going to ask... What about light mode? 😉

src/renderer/components/ClusterPage/clusterPageComponents.js
AntonNos: Removed it from `PanelsWrapper` component. Having scrollbar in parent component is better I think. Thanks

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: Great idea, thanks for tip! Done.

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: Oh, right. Sorry, I forgot about that case. Added it.
Thank you!

src/strings.ts
AntonNos: I asked @john-papadami about text strings on Tuesday, and he confirmed, that actual text strings are in Figma.
I leave them as it is for now. If it is necessary to change it, will do that. Just let me know what to do with text.
Thank you

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: Hmmm, you are right, we don't need that `if (cpuMetrics !== null)` and else block here. Thanks for tip

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: It checks if component was mounted. Found that here -> https://stackoverflow.com/questions/39767482/is-there-a-way-to-check-if-the-react-component-is-unmounted

stefcameron: OK, but what's the benefit of knowing this? So far, I'm not seeing how that helps anything. 🙂

src/strings.ts
stefcameron: I see. John and I didn't expressly "sync" our strings between the ticket and Figma.

My main concern with the strings describing the metrics is that they don't feel precise enough to differentiate them from what users already see in Lens in terms of CPU, Memory, and Storage. I think users will look at this and wonder why we put this in -- and worse, that we're probably getting it wrong because our metrics won't look like the ones in Lens.

But also, many of the strings, as they are now in the code, are missing words. They are incomplete. Either those are typos, or...? Like "Used: 1-minute average time spent performing." -- performing what? Or 
Capacity: Total disk space across." -- across what?

I think the strings in the ticket are more complete.

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: You are right. We don't need it:)
Thanks, removed

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: Fixed

src/strings.ts
AntonNos: Got it, will change them to be similar with ticket.

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: Great idea. It really works a lot faster that serially requests 🔥
Done

src/strings.ts
```diff
@@ -382,6 +382,72 @@ export const clusterPage: Dict = {
...
+            used: () => 'Used',
+            system: () => 'System',
+            io: () => 'I/O',
+            idle: () => 'Idle',
+            tooltipInfoHtml: () => `
+              <p>Metrics are retrieved from the kubernetes cluster and may differ from OS level metrics.</p>
+              <p>Used (%): 1-minute average time spent performing tasks, across the cluster.</p>
+              <p>System (%): 1-minute average time spent in the kernel, across the cluster.</p>
+              <p>I/O (%): 1-minute average time spent waiting for I/O, meaning system was disk- or network-bound, across the cluster.</p>
+              <p>Idle (%): 1-minute average idle time not performing tasks, across the cluster.</p>
```
stefcameron, line 407: I had a look at the tooltip in Lens. I think we could improve on it a bit to make it a little more clear by making a list out of the items, and then putting the disclaimed at the bottom with an asterisk (*) next to it, like this:

---

\- Used (%): 1-minute average time spent performing tasks, across the cluster.
\- System (%): 1-minute average time spent in the kernel, across the cluster.
\- I/O (%): 1-minute average time spent waiting for I/O, meaning system was disk- or network-bound, across the cluster.
\- Idle (%): 1-minute average idle time not performing tasks, across the cluster.

\* Metrics are retrieved from the kubernetes cluster and may differ from OS level metrics.

---

```
<p>- Used (%): 1-minute average time spent performing tasks, across the cluster.</p>
<p>- System (%): 1-minute average time spent in the kernel, across the cluster.</p>
<p>- I/O (%): 1-minute average time spent waiting for I/O, meaning system was disk- or network-bound, across the cluster.</p>
<p>- Idle (%): 1-minute average idle time not performing tasks, across the cluster.</p>
<p class="disclaimer">* Metrics are retrieved from the kubernetes cluster and may differ from OS level metrics.</p>
```

And then style the `p.disclaimer` class in MetricTitle.js's `<Title>` styled component to have an 8px top margin so it looks like this:

image

Same for the memory and storage tooltips.

stefcameron:
> > I made 3 fullwidth columns, because only on >1920px monitors it looks good with 50% columns.
> 
> Really? It's not possible to make the graphs smaller (and so narrower) to give in a side-by-side panel with Overview?

Right now, I can shrink the Lens window (making it narrower) and the health panel responds, including the graphs:

image

The graphs still look good. Why can't we have the Health panel 50% on the right, with Overview 50% on the left, like in the mockups?

stefcameron (commented):
Looking good! I just have a small request to tweak the tooltips, and then still wondering why the health panel can't be narrower -- see my previous comment in the PR.

src/strings.ts
AntonNos: Nice idea, thank you for tip about tooltip😄
Fixed. Also fixed width of columns in Overview panel.

stefcameron (commented):
Visually, all the changes look great now!

But, I discovered that if I switch from dark to light mode, or vice versa, the background color of the charts doesn't switch.

For example, when going from dark to light:

Screenshot 2023-01-17 at 5 29 14 PM
Screenshot 2023-01-17 at 5 29 35 PM

The same problem occurs if you start out in light mode and then change to dark mode. The background color will remain the one in light mode.

AntonNos:
> Visually, all the changes look great now!
> 
> But, I discovered that if I switch from dark to light mode, or vice versa, the background color of the charts doesn't switch.
> 
> For example, when going from dark to light:
> 
> Screenshot 2023-01-17 at 5 29 14 PM
> 
> Screenshot 2023-01-17 at 5 29 35 PM
> 
> The same problem occurs if you start out in light mode and then change to dark mode. The background color will remain the one in light mode.

Hmm, didn't see that. 
Fixed it by adding `theme.mode` to `useEffect` dependencies.

src/renderer/components/ClusterPage/Overview/MetricChart.js
```diff
@@ -63,10 +64,11 @@ const options = {
 
 export const MetricChart = ({ chartColor, chartFillPercentage }) => {
   const [chartData, setChartData] = useState(getDefaultData());
+  const { theme } = useThemeSwitching();
 
   useEffect(() => {
     setChartData(getData(chartColor, chartFillPercentage));
-  }, [chartColor, chartFillPercentage]);
+  }, [chartColor, chartFillPercentage, theme.mode]);
```
stefcameron, line 71: In theory, this seems good, but when I run it, while the colors to update, the chart data doesn't load (or remains at zero, or gets reset to zero, not sure):

image

AntonNos: Fixed. Added new error message and reconnect button.

src/strings.ts
```diff
@@ -385,13 +385,20 @@ export const clusterPage: Dict = {
...
-              'Health metrics cannot be displayed for this cluster for one of the following reasons:',
-            reasonsList: () => [
-              'Stacklight is disabled.',
-              'Stacklight service is not available.',
-            ],
+          error: {
+            disconnectedManagementCluster: {
+              title: () =>
+                'Health metrics cannot be displayed for this cluster because its management cluster is disconnected.',
+              reconnectButtonLabel: () => 'Reconnect!',
```
stefcameron, line 392: I think just "Reconnect" without the exclamation mark (!) is sufficient.

stefcameron (changes requested):
Thank you for the update! In general, this looks good, but it's not really working well.

When disconnected, I see this:

Screenshot 2023-01-20 at 9 29 23 AM

The paragraph should be up next to the error icon, and the Reconnect link should be right at the end of the text, not on the next line.

Also, when I click Reconnect, the spinners are a nice touch, but they never go away. I stays like this forever:

Screenshot 2023-01-20 at 9 30 43 AM

src/strings.ts
AntonNos: Fixed

AntonNos:
> Thank you for the update! In general, this looks good, but it's not really working well.
> 
> When disconnected, I see this:
> 
> Screenshot 2023-01-20 at 9 29 23 AM
> 
> The paragraph should be up next to the error icon, and the Reconnect link should be right at the end of the text, not on the next line.
> 
> Also, when I click Reconnect, the spinners are a nice touch, but they never go away. I stays like this forever:
> 
> Screenshot 2023-01-20 at 9 30 43 AM

Fixed. Now we have a spinner while `isCloudFetching = true` and once it updates, charts and info updates too

stefcameron (changes requested):
Thanks! It's better, but there are still some issues:

Screenshot 2023-01-23 at 7 19 05 PM

1. The text shouldn't go under the error icon. Usually this means a flex container with an icon and a paragraph, and the icon uses `flex: none` and paragraph `flex: 1` (so it can grow/shrink). That should result in the paragraph not flowing under the icon.
2. The CPU, Memory, and Storage labels don't appear centered within their columns with respect to their graphs.
3. I somehow got into this state where the bottom of the health panel was missing its gap. Not sure how. Can't always repro.

Finally, there's what we call "jank" when reconnecting (see how it flashes back to "disconnected" error and back to spinners while it's reconnecting):

cclex-jank

I think you need a local state variable to put yourself into spinner mode as soon as the link is clicked rather than relying solely on the cloud events.

Just make sure to come back out of it if the reconnection fails... We should end up back at this disconnected error if reconnection fails.

stefcameron:
@AntonNos About

> The text shouldn't go under the error icon. Usually this means a flex container with an icon and a paragraph, and the icon uses flex: none and paragraph flex: 1 (so it can grow/shrink). That should result in the paragraph not flowing under the icon.

Maybe you can use our little `<InlineNotice>` component...?

AntonNos:
> @AntonNos About
> 
> > The text shouldn't go under the error icon. Usually this means a flex container with an icon and a paragraph, and the icon uses flex: none and paragraph flex: 1 (so it can grow/shrink). That should result in the paragraph not flowing under the icon.
> 
> Maybe you can use our little `<InlineNotice>` component...?

Thanks for tip! I will update PR today and will try to use `<InlineNotice>` here.

AntonNos:
> Thanks! It's better, but there are still some issues:
> 
> Screenshot 2023-01-23 at 7 19 05 PM
> 
> 1. The text shouldn't go under the error icon. Usually this means a flex container with an icon and a paragraph, and the icon uses `flex: none` and paragraph `flex: 1` (so it can grow/shrink). That should result in the paragraph not flowing under the icon.
> 2. The CPU, Memory, and Storage labels don't appear centered within their columns with respect to their graphs.
> 3. I somehow got into this state where the bottom of the health panel was missing its gap. Not sure how. Can't always repro.
> 
> Finally, there's what we call "jank" when reconnecting (see how it flashes back to "disconnected" error and back to spinners while it's reconnecting):
> 
> cclex-jank [ cclex-jank ](https://user-images.githubusercontent.com/2855350/214194089-74949eb1-cd5f-499e-9077-03dad12ec4e0.gif) [ ](https://user-images.githubusercontent.com/2855350/214194089-74949eb1-cd5f-499e-9077-03dad12ec4e0.gif)
> 
> I think you need a local state variable to put yourself into spinner mode as soon as the link is clicked rather than relying solely on the cloud events.
> 
> Just make sure to come back out of it if the reconnection fails... We should end up back at this disconnected error if reconnection fails.

Fixed! Now it works much faster than before. Also used `InlineNotice` component for warning message, fixed bug with gap and centered title with tooltip icon.

stefcameron (changes requested):
Thank you for the update!

We're really, really close now. But my eye, it just keeps see things. 😉 

It bugs me that the error icon isn't centered with the first line of text:

Screenshot 2023-01-25 at 7 08 18 PM

All that's needed here is to fix `InlineNotice.js` like this at line 34:

```
const Container = styled.div(function () {
  return {
    display: 'flex',

    '> .Icon': {
      marginTop: 2,
    }
  };
});
```

(i.e. move the icon down 2px)

Then we get this (which also works with the other icons related to the `type` prop, like WARNING, INFO, NOTE, etc.):

Screenshot 2023-01-25 at 7 10 02 PM

And last but not least, if the Cloud is disconnected, but the cluster was already synced, and the cluster's `lma` property is defined, but for some reason, `lma.prometheus` is not (i.e. StackLight is there, but not Prom, for some reason, not sure if it's possible, but when dealing with MCC-related things, I not to make too many assumptions...), then I get this:

Screenshot 2023-01-25 at 7 11 53 PM

Both errors show-up at the same time. The StackLight error should take precedence over the disconnected error because even if we were reconnected, we still wouldn't have data.

stefcameron:
> The StackLight error should take precedence over the disconnected error because even if we were reconnected, we still wouldn't have data.

With your fix, when we have both conditions as the same time (disconnected and no Prom URL), we see the "disconnected" error message -- but what we should see is the "no StackLight" error message, because even if they reconnected, we still wouldn't be able to show anything...

If they're disconnected _with_ a Prom URL, then we should see the "disconnected" message.

Makes sense?

AntonNos:
> > The StackLight error should take precedence over the disconnected error because even if we were reconnected, we still wouldn't have data.
> 
> With your fix, when we have both conditions as the same time (disconnected and no Prom URL), we see the "disconnected" error message -- but what we should see is the "no StackLight" error message, because even if they reconnected, we still wouldn't be able to show anything...
> 
> If they're disconnected _with_ a Prom URL, then we should see the "disconnected" message.
> 
> Makes sense?

I'm showing a message with list of problems without "Reconnect" button. Is that wrong message for that case (in both conditions in the same time)?

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -350,6 +350,21 @@ export const HealthPanel = ({ clusterEntity }) => {
       <DrawerTitleWrapper>
         <PanelTitle title={strings.clusterPage.pages.overview.health.title()} />
       </DrawerTitleWrapper>
+      {(isNoMetrics && cloudStatus === CONNECTION_STATUSES.DISCONNECTED) ||
+        (isNoMetrics && (
```
stefcameron, line 354: The logic is a bit hard to understand here because there seem to be overlapping conditions related to `isNoMetrics === true`.

Feels like it should be `{isNoMetrics ? <ErrorMessage ... /> : null}` (BTW, it's a better practice to use the ternary operator and return `null` instead of doing `isNoMetrics && <ErrorMessage ...` because the former really results in nothing in the DOM while the latter `&&` case still affects the DOM)

And then line 368 should be `{!isNoMetrics && cloudStatus === CONNECTION_STATUSES.DISCONNECTED ? ... : null}`

BTW, double-negatives are really hard to process in the human brain and you should always try to avoid them as much as possible.

It's always better to find a positive way to name variables so you don't end up with cases like `!isNoMetrics` -- what? Oh, "not-no-metrics", so we _have_ metrics. That's a lot harder to process than `!haveMetrics` or `!promUrl` -- "I don't have metrics" or "No Prom URL". 🙂 

Like I wish HTML had never come up with a `disabled` attribute! That's a double-negative, because you inevitably come up to cases where you have to test, `if (!disabled)` -- much harder to reason in your brain than, `if (enabled)`, right?

stefcameron (changes requested):

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: You are right, bad naming is hard to understand. I'm trying to name variables logically, in that case I couldn't think of a better name that "isNoMetrics"🤷‍♂️
Will take into account about ternary operator instead of `... && ()`.
Thank you for tips!
I'll make the changes in a few hours, because I'm not near computer right now.

stefcameron:
> 

This is what I see when I'm both disconnected _and_ don't have prometheus on a cluster:

image

I should be seeing the other message instead.

AntonNos:
> > > > The StackLight error should take precedence over the disconnected error because even if we were reconnected, we still wouldn't have data.
> > > 
> > > 
> > > With your fix, when we have both conditions as the same time (disconnected and no Prom URL), we see the "disconnected" error message -- but what we should see is the "no StackLight" error message, because even if they reconnected, we still wouldn't be able to show anything...
> > > If they're disconnected _with_ a Prom URL, then we should see the "disconnected" message.
> > > Makes sense?
> > 
> > 
> > I'm showing a message with list of problems without "Reconnect" button. Is that wrong message for that case (in both conditions in the same time)?
> 
> This is what I see when I'm both disconnected _and_ don't have prometheus on a cluster:
> 
> image
> 
> I should be seeing the other message instead.

Fixed, now we have correct message. I think the problem was because of `... && ()`. Now it works good. Thanks again for tip!

src/renderer/components/ClusterPage/Overview/HealthPanel.js
```diff
@@ -350,22 +350,22 @@ export const HealthPanel = ({ clusterEntity }) => {
...
-                .reasonsList()
-                .map((reason, index) => (
-                  <li key={index}>{reason}</li>
-                ))}
-            </ol>
-          </ErrorMessage>
-        ))}
-      {cloudStatus === CONNECTION_STATUSES.DISCONNECTED && (
+      {(!hasPromUrl && cloudStatus === CONNECTION_STATUSES.DISCONNECTED) ||
+      !hasPromUrl ? (
```
stefcameron, line 354: Because of the `&&` operator here, and the fact that on both sides of the `||` you're testing the same condition `!hasPromUrl`, the entire expression can be reduced to just `!hasPromUrl`.

If we don't have a prom URL, then the connection status really doesn't matter: We need to show we can't get the data because the StackLight config doesn't support it.

And then line 368 is good because it read, "if we do have a prom url, and we're disconnected, then show the notice about having no data because we're disconnected and offer to reconnect."

I tested the above locally to be 100% sure.

stefcameron (changes requested):
That's it. The last thing.

src/renderer/components/ClusterPage/Overview/HealthPanel.js
AntonNos: You are right, we don't care about connection status if we don't have a prom URL. Fixed

stefcameron (approved):
🚀 Thank you for sticking with it to the end and getting all those smaller issues fixed too!

AntonNos:
> 🚀 Thank you for sticking with it to the end and getting all those smaller issues fixed too!

Thank you for all the tips and comments!

